PR #16, opened 2018-11-07
Fix typo in todo-list docs

deweller:


moxiegirl:
@deweller Thanks for the catch and the PR!

PR #17, opened 2018-11-07
Replace React with Vue

deweller:
Replace references to React with Vue.js


moxiegirl:
@deweller Thanks.  Missed that in cut and paste. ✂️

PR #18, opened 2018-11-07
Use npm instead of yarn

deweller:
We can use npm instead of yarn.

While yarn may be preferable for advanced users, it does create an additional external dependency.  For a beginner tutorial, I suggest npm is preferable.

Modern npm has lockfile capability and is arguably as performant as yarn now.

### DIFF
--- a/_browser/todo-list.md
+++ b/_browser/todo-list.md
@@ -67,18 +67,13 @@ These instructions assume you are cloning.
     $ cd blockstack-todos
     ```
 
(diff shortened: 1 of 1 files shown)


moxiegirl:
@deweller Let me check with the team on this one. If I recall, this point was made before but there was some reason they wanted to keep yarn.

moxiegirl:
@deweller Ok. Looks like we are going to do this. Thanks again for your help.

PR #56, opened 2018-12-14
Fix whitepaper links

friedger:
Use the correct title for white papers.

Fixes #56


moxiegirl:
@friedger Sorry. Just saw this. I fixed this on another PR after seeing your issue. Thanks for the request.

PR #59, opened 2018-12-18
fix typos

CoralSilver:


moxiegirl:
@CoralSilver Thanks for the catches and the PR

PR #84, opened 2018-12-26
Fix typos

huyhong:


moxiegirl:
@huyhong Thank you for the fixes!

Issue #82, opened 2018-12-22
Expose GAIA repo docs in the documentation site


moxiegirl:
Currently, we are pointing users to the GAIA repo documentation. We want to expose that documentation to customers and to search.

markmhx:
This work includes providing instructions for setting up Gaia hubs with:

- Docker compose (Virginia's work)
- Digital Ocean (Hank's work)

moxiegirl:
Ok, this is its own issue and so is #80 @markmhx --- they are two blocks of work.

PR #111, opened 2019-01-24
[REVIEW]  Gaia move to docs.blockstack

moxiegirl:
- Stand up a Gaia section in docs.blockstack.org
- Add in the how to deploy on Digital Ocean #82
- Cover the AWS AMI + Compose


_storage/hub-operation.md
```diff
@@ -0,0 +1,108 @@
...
+the hub server. Alternatively, you can specify a file location using the
+`CONFIG_PATH` environment variable.  The following is an example configuration file for Amazon S3:
+
+```json
+{
+  "servername": "localhost",
+  "port": 4000,
+  "driver": "aws",
+  "readURL": "https://YOUR_BUCKET_NAME.s3.amazonaws.com/",
+  "proofsConfig": {
```
kantai: I'd just leave `proofsConfig` out of the sample entirely. The config file doesn't need to have it, and we should deprecate this proof checking feature in any case.

_storage/hub-operation.md
```diff
@@ -0,0 +1,108 @@
...
+}
+```
+
+### The readURL parameter
+
+By default, hub drivers return read URLs that point directly at the written content. For example, an S3 driver would return the URL directly to the S3 file. If you configure a CDN or domain to point at that same bucket, you can use the `readURL` parameter to tell the hub that files can be read from a given URL. For example, the `hub.blockstack.org` Gaia Hub is configured to return a read URL that looks like `https://gaia.blockstack.org/hub/`.
+
+Unset the `readURL` parameter if you do not intend to deploy any caching.
+
+### The proofsRequired parameter
```
kantai: Can we add a note here that this feature is deprecated? I'll add a note to the Gaia readme as well

_storage/hub-operation.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Open-membership hub
+
+An open-membership storage hub permits writes for _any_ address top-level
+directory. Every request is validated such that write requests must provide
+valid authentication tokens for that address. Operating in this mode is
+recommended for service and identity providers who wish to support many
+different users.
+
+To limit the users that may interact with such a hub, you can limit users to
```
kantai: Similarly -- I'd prefer if we didn't talk about social proofs, because I'm not sure that this feature works at the moment.

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,372 @@
...
+            "mode": 420,
+            "contents": {
+              "source": "data:application/octet-stream,DOMAIN%3D<DOMAIN_NAME_VALUE>%0ASTAGING%3D<STAGING_VALUE>"
+            }
+          }]
+        }
+      }
+      ```
+
+      A value of `0` to requests a valid SSL certificate. At this point, you'll
```
kantai: it's not clear at this point what the `0` is referring to -- I think we could just remove this sentence, since instructions are given in step 6.

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,372 @@
+---
+layout: storage
+permalink: /:collection/:path.html
+---
+# Configure a hub on Amazon S3
```
kantai: These instructions read fine to me -- I'll need to test them once I've synced with Jesse on the AMI status

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,372 @@
...
+
+   For example, you can use the prefix `www` with your domain name. When you are done, your
+
+6. Save your changes.
+
+   At this point, your DNS management should look similar to the following except that with your domain rather than the `maryhub.ga` domain.
+
+   ![DNS fields](/storage/images/aws-dns-fields.png)
+
+7. After your changes propagate, visit your new domain at the `hub_info` page.
```
kantai: For me -- this sent me to https://my-domain-name -- which triggered a "Your connection is not private" page. We might want to add a note here.

_data/navigation_storage.yml
```diff
@@ -0,0 +1,18 @@
+- title: Introduction
```
markmhx: Might it be clearer to title this "Introduction to storage with Gaia" since it's not at first glance very obvious what category / type of the docs the user is viewing otherwise?

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
+---
+layout: storage
+permalink: /:collection/:path.html
+---
+# A decentralized storage architecture
+
+The Blockstack Network stores application data using a storage service called
+Gaia. Transactional metadata is stored on the Blockstack blockchain and user
+application data is stored in Gaia hub. Storing data off of the blockchain
```
markmhx: "Stored in **a** Gaia hub" or perhaps even better "stored in a network of Gaia hubs"?

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
+---
+layout: storage
+permalink: /:collection/:path.html
+---
+# A decentralized storage architecture
+
+The Blockstack Network stores application data using a storage service called
```
markmhx: Do we want to call Gaia a "system" rather than a "service" since the latter may connote a centralized server-based resource? 

We seem to use the word "system" in [Gaia's top-level README](https://github.com/blockstack/gaia) with "This document describes the high-level design and implementation of the Gaia storage system".

_includes/architecture.md
```diff
@@ -0,0 +1,21 @@
+
+![Blockstack Architecture](/common/images/architecture.png)
+
+Blockchains require consensus among large numbers of people, so they can be slow. Additionally, a blockchain is not designed to hold a lot of data. This means using a blockchain for every bit of data a user might write and store is expensive. For example, image if an application were storing every tweet in the chain.
+
+Blockstack addresses blockchain performance problems by taking a layered approach. At the base of the system is a blockchain and the Blockstack Naming System(BNS). The blockchain  governs ownership of names (identities)  in the system, names such as domain names, username, and applications names.
```
markmhx: Needs space before "(BNS)"

_includes/architecture.md
```diff
@@ -0,0 +1,21 @@
...
+
+Names in Blockstack correspond to routing data in the OSI stack. The routing data is stored in the Atlas Peer Network, the second layer. Every core node that joins the Blockstack Network is able to obtain an entire copy of this routing data. Blockstack uses the routing data to associate names (usernames, domains, application names) with a particular storage location.
```
markmhx: Should we write out "OSI" in full since it's the first usage on the page? I'm not personally sure what it is.

_includes/architecture.md
```diff
@@ -0,0 +1,21 @@
+
+![Blockstack Architecture](/common/images/architecture.png)
+
+Blockchains require consensus among large numbers of people, so they can be slow. Additionally, a blockchain is not designed to hold a lot of data. This means using a blockchain for every bit of data a user might write and store is expensive. For example, image if an application were storing every tweet in the chain.
```
markmhx: "imagine" instead of "image"

_includes/architecture.md
markmhx: Also "application names" not "applications" plural

markmhx: "usernames" instead of "username singular

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+Gaia. Transactional metadata is stored on the Blockstack blockchain and user
+application data is stored in Gaia hub. Storing data off of the blockchain
+ensures that Blockstack applications can provide users with high-performance and
```
markmhx: "high performance" sans hyphen

_includes/architecture.md
markmhx: Also "(usernames, domains, and application names)" with "and"

_includes/architecture.md
```diff
@@ -0,0 +1,21 @@
...
+
+Blockstack addresses blockchain performance problems by taking a layered approach. At the base of the system is a blockchain and the Blockstack Naming System(BNS). The blockchain  governs ownership of names (identities)  in the system, names such as domain names, username, and applications names.
+
+Names in Blockstack correspond to routing data in the OSI stack. The routing data is stored in the Atlas Peer Network, the second layer. Every core node that joins the Blockstack Network is able to obtain an entire copy of this routing data. Blockstack uses the routing data to associate names (usernames, domains, application names) with a particular storage location.
+
+The final layer is the Gaia Storage System. Gaia works by hosting a _Gaia hub_ in one or more existing storage systems.
```
markmhx: The pink color for italics is making me think they're links. I wonder if giving italics a slightly colored background highlight would prevent this? Or some other style treatment?

markmhx: I'm not sure here what a "storage system" is, especially in relation to a "storage provider" below.

My understanding is that Gaia hubs basically serve to connect applications with storage provider on behalf of users.

Do you mean to refer to a server host (e.g. DigitalOcean, EC2) as the "storage system" where a hub is installed? I wonder if there's a clearer way to lay out these pieces and how they connect to each other.

_includes/architecture.md
```diff
@@ -0,0 +1,21 @@
...
+These storage systems typically belong to storage providers such as Dropbox,
+Amazon EC2, and so forth. Blockstack currently has driver support for S3 and
```
markmhx: Should we say that "Gaia" not "Blockstack" has driver support for these, and even before that, say that "drivers" are software components for Gaia hubs that permit them to connect to specific storage providers?

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+
+##  Understand Gaia in the Blockstack architecture
+
+The following diagram depicts the Blockstack architecture and Gaia's place in it:
+
+{% include architecture.md %}
+
+## User control or how is Gaia decentralized?
+
+Within the provider's storage system a Gaia hub runs as a service. This service
```
markmhx: Similar to my note above, I'm not sure what a storage system is, let alone a "provider's" storage system.

markmhx: This sentence that starts with "This service authenticates writes to a location" doesn't make a lot of sense to me, can you try rephrasing?

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+
+Gaia's approach to decentralization focuses on user control of data and storage.
+If a user can choose which storage provider to store data with, then that is all the decentralization required to enable user-controlled
+applications.
+
+The control of user data lies in the way that user data is accessed.
+When an application fetches a file `data.txt` for a given user `alice.id`, the
+lookup will follow these steps:
+
+1. Fetch the `zonefile` for `alice.id`
```
markmhx: Should we mention "...via BNS"?

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+2. Read her profile URL from her `zonefile`.
+3. Fetch Alice's profile.
```
markmhx: Should we add "...as served by a Gaia hub at the profile URL?" or something of that sort?

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+4.  _Verify_ that the profile is signed by `alice.id`'s key
+5. Read the application root URL (e.g. `https://gaia.alice.org/`) out of the profile
+6. Fetch file from `https://gaia.alice.org/data.txt`.
+
+Because `alice.id` controls her zonefile, she can change where her profile is
```
markmhx: "Because `alice.id` controls her zonefile **with her Blockstack ID**"?

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+stored. For example, she may do this if the current storage of the profile is
+compromised. Similarly, if Alice wishes to change her Gaia provider, or run her
+own Gaia node, she can change the entry in her profile.
```
markmhx: "Similarly, if Alice wishes to change the storage provider she uses to store her application data with Gaia, or run her own Gaia hub, she can change the entry in her profile"?

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+
+Applications writing directly on behalf of Alice, do not need to perform this
```
markmhx: I think the comma here can be removed

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Understand data storage
+
+A Gaia storage hub stores the written data _exactly_ as given. It offers minimal
+guarantees about the data. It does not ensure that data is validly formatted,
+contains valid signatures, or is encrypted. Rather, the design philosophy is
+that these concerns are client-side concerns.
+
+Client libraries (such as `blockstack.js`) are capable of providing these
+guarantees. Blockstack used a liberal definition of the end-to-end principle to
```
markmhx: Maybe worth linking to a page describing this principle for those unfamiliar? e.g. https://en.wikipedia.org/wiki/End-to-end_principle

_storage/authentication.md
```diff
@@ -0,0 +1,31 @@
+---
+layout: storage
+permalink: /:collection/:path.html
+---
+# Authentication and storage
+
+Blockstack authentication is a bearer token-based authentication system. From an app user's perspective, login similar to third-party authentication techniques that they're familiar with. For an app developer, the flow is unlike the typical client-server flow of centralized sign-in services such as OAuth. With Blockstack the authentication flow happens entirely client-side.
```
markmhx: I think this is sentence is missing a word?

markmhx: "bearer-token-based"

_storage/authentication.md
```diff
@@ -0,0 +1,31 @@
...
+
+In this section, you get an overview of the authentication system and learn how Gaia fits into it.
+
+## Authentication and Gaia
+
+Blockstack authentication is a bearer token-based authentication system. From an app user's perspective, login similar to third-party authentication techniques that they're familiar with. For an app developer, the flow is unlike the typical client-server flow of centralized sign-in services such as OAuth. With Blockstack the authentication flow happens entirely client-side.
```
markmhx: Is this paragraph redundant with the above section?

_storage/authentication.md
```diff
@@ -0,0 +1,31 @@
...
+
+A decentralized application (DApp) and the Blockstack Browser communicate during the authentication flow by passing back and forth two tokens. The requesting application sends the Blockstack Browser an authRequest token. Once a user approves a sign-in, the Blockstack Browser responds to the application with an `authResponse` token. These tokens are JSON Web Tokens, and they are passed via URL query strings.
```
markmhx: Does `authRequest` here need to be highlighted the same as with `authResponse`?

markmhx: Also should we link "JSON Web Tokens" to https://jwt.io/ ?

_storage/authentication.md
```diff
@@ -0,0 +1,31 @@
...
+
+When a user chooses to "Sign in with Blockstack" on your DApp, the `redirectToSignIn()` method sends the user to the Blockstack Browser. The browser responds with an authentication token and an _app private key_.
+
+![](/storage/images/app-sign-in.png)
+
+The app private key is application -specific. It is generated from the user's identity address private key using the `appDomain` as input. This key is deterministic, meaning that for a given Blockstack ID and domain_name, the same private key is generated each time. The app private key is securely shared with the app on each authentication and encrypted by the Blockstack Browser. The key serves three functions, it:
```
markmhx: "application-specific" sans space

markmhx: "domain name" instead of "domain_name" ?

_storage/authentication.md
```diff
@@ -0,0 +1,31 @@
...
+
+* is used to create the credentials that give an app access to the Gaia hub storage bucket for that specific app
+* is used in the end-to-end encryption of files stored for the app on the user's Gaia hub
+* serves as a cryptographic secret that apps can use to perform other cryptographic functions
+
+When an application writes to a Gaia hub, the authentication token, key, and the data are passed to the Gaia hub.
+
+![Gai writes](/storage/images/gaia-writes.png)
```
markmhx: "Gaia writes"

_storage/overview.md
```diff
@@ -0,0 +1,204 @@
...
+parties.
+
+* TOC
+{:toc}
+
+##  Understand Gaia in the Blockstack architecture
+
+The following diagram depicts the Blockstack architecture and Gaia's place in it:
+
+{% include architecture.md %}
```
markmhx: I think you grabbed this image from elsewhere, but FYI there's a typo with "Google Drive" in case we can fix

_storage/write-to-read.md
```diff
@@ -0,0 +1,56 @@
+---
+layout: storage
+permalink: /:collection/:path.html
+---
+# Storage write and read
+
+Once a user authenticates and a DApp obtains authentication, the application interacts with Gai through the blockstack.js library. There are two simple methods for working with data in Gaia hub the `putFile()` and `getFile()` methods. This section goes into greater detail about the methods, how they interact with a hub, and how to use them.
```
markmhx: "Gaia" not "Gai"

markmhx: Should we link "blockstack.js library" to the GitHub repo or other docs about it?

markmhx: "There are two simple methods for working with data in a Gaia hub: the `putFile()` and `getFile()` methods."

_storage/write-to-read.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+## Write-to and Read-from URL Guarantees
+
+Gaia is built on a driver model that supports many storage services. So, with very few lines of code, you can interact with providers on Amazon S3, Drop, and so forth.  The simple getFile and putFile interfaces are kept simple because Blockstack assumes and wants to encourage a community of OS data management libraries.
```
markmhx: "storage providers" instead of "services"?

markmhx: "Dropbox" instead of "Drop"

markmhx: Should we always highlight method names with double back hashes aka `getFile` instead of getFile here?

markmhx: "open-source data-management libraries"

_storage/write-to-read.md
```diff
@@ -0,0 +1,56 @@
...
+
+The performance and simplicity oriented guarantee of the Gaia specification is
```
markmhx: "simplicity-oriented"

_storage/write-to-read.md
```diff
@@ -0,0 +1,56 @@
...
+that when an application submits a write-to the
```
markmhx: `write-to` and later `read-from` highlighted

_storage/write-to-read.md
```diff
@@ -0,0 +1,56 @@
...
+`https://myhub.service.org/store/foo/bar` URL, the application is guaranteed to
+be able to read from the `https://myreads.com/foo/bar` URL. Note that, while the prefix in the write-to url (`myhub.service.org/store`) and the read-from URL (`https://myreads.com`) are different, the `foo/bar` suffixes are the same.
```
markmhx: "(e.g. `myhub.service.org/store`)" with the "e.g."

_storage/write-to-read.md
```diff
@@ -0,0 +1,56 @@
...
+https://myservice.org/read/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/0/profile.json
+```
+
+The application is guaranteed that the profile is written with `putFile()` this request address:
+
+```
+https://myservice.org/store/1DHvWDj834zPAkwMhpXdYbCYh4PomwQfzz/0/profile.json
+```
+
+When you use the `putFile()` method it takes the user data and POSTs it to the user's Gaia storage hub. The data POSTs directly to the hub, the blockchain is not used and no data is stored there.
```
markmhx: Should we explain somewhere around here why there isn't a `write_url_prefix` returned by the `hub_info` endpoint like we have a `read_url_prefix` available?

_storage/write-to-read.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+## Address-based access-control
+
+Access control in a Gaia storage hub is performed on a per-address basis.
+Writes to URLs `/store/<address>/<file>` are only allowed if the writer can
```
markmhx: "allowed only"

_storage/write-to-read.md
```diff
@@ -0,0 +1,56 @@
...
+demonstrate that they control _that_ address. This is achieved via the
+authentication token which is a message _signed_ by the private key associated
+with that address. The message itself is a challenge-text, returned via the
```
markmhx: "challenge text" sans hyphen?

_storage/hub-choice.md
```diff
@@ -0,0 +1,138 @@
...
+---
+# Hubs and user choice
+
+Blockstack operates a default Gaia storage hub at `https://hub.blockstack.org`.
+Individuals or organizations may also run their own hubs, either as a for-profit
+service or for other reasons.
+
+## Background
+
+Each user with an identity in the Blockstack Network has a storage hub
```
markmhx: "Gaia hub"?

_storage/hub-choice.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Enable the Gaia hub URL option for new users
+
+A major principle of Blockstack applications is that DApps do not store or
+replicated user data. Through the Gaia feature, Blockstack seeks to empower
+users to choose where to store their data and how they manage it.
+
+As a step towards fulfilling this principle, DApp developers can configure their
+application to prompt new users to provide a Gaia hub URL. An application with
+this feature enabled, presents users creating an identity for the first time
```
markmhx: sans comma

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,466 @@
...
+
+1. Visit the <a href="https://aws.amazon.com/free/" target="\_blank">AWS Free Tier page</a> and choose **Sign in to the Console**.
+
+   ![](/storage/images/aws-console.png)
+
+2. Make sure your region is set to `us-west-2` which is **Oregon**.
+
+   ![](/storage/images/us-west-2.png)
+
+3. Choose **Launch a virtual machine**.
```
markmhx: My AWS UI seems to say "Launch Instance"s with a blue button instead of this

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,466 @@
...
+
+   5. Replace `<KEYPHRSE>` with a key phrase.
+
+         For example, `hubbadubbado` is a fun key phrase.
+
+   6. Replace `<DOMAIN_NAME_VALUE>` with your domain name.
+
+      For example, `maryhub.ml` is my domain name.
+
+   7. Replace the `<STAGING_VALUE>` with the number `0`.
```
markmhx: If we're asking them to replace this with a constant value of `0`, should we just place that directly into the blob above?

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,466 @@
...
+    You should see a response from your Gaia hub!
+
+    ![Hub test](/storage/images/aws-hub.png)
+
+    At this point, you should see a **Not secure** message in the browser. That's because you haven't yet enabled SSL certification.
+
+
+## Task 3: Configure a domain name
+
+At this point, you can point a domain to your Gaia hub. Although it's not required, it is highly recommended. If you use a domain, you can migrate your instance to a different server (or even provider such as Azure or Dropbox) at any time, and still access it through the domain URL. Just point your domain at the IP address for your EC2 instance. Use an `A Record` DNS type.
```
markmhx: It may be worth mentioning that while HTTPS may not required to simply have a hub running via HTTP, it _is_ required to connect the hub to your Blockstack ID for usage. 

i.e. the Blockstack UI will reject any non-HTTPs URLs that users attempt to set for configuration

_storage/hub-operation.md
moxiegirl: Sure. Removed mention from all Hub docs.

_storage/amazon-s3-deploy.md
moxiegirl: Fixed.

_storage/amazon-s3-deploy.md
moxiegirl: Cool. Thanks.

_data/navigation_storage.yml
moxiegirl: ok.

_storage/overview.md
moxiegirl: So, there is only a single hub where user data is stored. In terms of implementing the service, the designer might choose to set up several and load balance among them.  That is a service implementation detail.  Stet.

_storage/overview.md
moxiegirl: Corrected per group discussion.

_includes/architecture.md
moxiegirl: OSI is the standard 7 layer model describing a network from physical to the application layer.  Network people pretty much know it by heart.  So, routing is a layer 3 activity.

_includes/architecture.md
moxiegirl: nice catch

_includes/architecture.md
moxiegirl: I played around with the pink color. I think it is something we can look into.  For now, I'll leave that part unchanged.  I'll look at tweaking style after it goes out.

Gaia strictly literally provides a hub service and data storage to a user.  Gaia runs as a service on a compute resource (machine) and Gaia typically writes to disk space (storage) that also belong to the same storage provider.  In theory, you could configure the Gaia hub service on one provider and the storage space on an entirely different provider.  Network resources don't care about the physical location they reference connection. (Oversimplification, this leaves aside deeper questions of performance, latency etc).

Further, the use of containers and domain names for a Gaia service configuration means any entity (person or org) running a Gaia hub can switch a Gaia service and storage from one storage provider to another. Technically it is a matter of moving a container and copying disk space (not the same as migration). Since we can envision a world where people offer Gaia hubs to users, these Gaia hub providers will want the flexibility of moving their service from one compute/storage provider to another depending on pricing, performance, or other considerations generally invisible to users of a Gaia hub.

While there may be users that care which storage provider a Gaia hub is on,
It is confusing to use the term "storage provider" except in the sense it is most commonly used. "Cloud storage provider" or "storage provider" for short is terminology (123,000,000 results) meaning companies that offer storage _and also compute resources_.  Storage provider we should reserve for Azure, Dropbox, etc.

We can use "hub provider" or "hub service" for Gaia.  This is why I was complaining about the use of the words "storage provider" in relation to Gaia interfaces. The phrase is confusing. We should say "storage hub" or "hub provider" or "storage hub URL" or "Gaia storage hub."  The hub should be immaterial to us.  The actual provider could become a point of sale _for the Gaia hub provider_ at some point but those services are likely to become mere commodity even to the person running the hub.

@kantai Just a check here please if I've misunderstood.

moxiegirl: "data hub" or "data storage URL" actually incredibly more user friendly when you think about it.

_includes/architecture.md
moxiegirl: Yep.  A Gaia configuration enables the connection. A storage driver is super common concept to network admins -- generally analogous to a printer driver.  Admins generally don't have to concern themselves with the driver except to know it exists, they'll know this through the config options.

_storage/overview.md
moxiegirl: I've tweaked the language on the architecture section in light of this and your previous comments.

_storage/overview.md
moxiegirl: Not at this point; it would shift the focus.

_storage/overview.md
moxiegirl: At step 3, it doesn't know where Gaia lives. I don't believe the profile is on the Gaia hub.  The profile is defined with the identity.  Both can be gotten by querying the blockchain explorer. 

@kantai is the profile actually in Gaia storage?

_storage/overview.md
moxiegirl: `alice.id` = the Blockstack ID

`Because alice.id controls her zone file with alice.id` doesn't actually make sense. 

Alice controls her zonefile by virtue of the fact that she has the secret key for it.

`Because `alice.id` has access to her zonefile, she can change where her profile
is stored. `

Remember, we don't actually have a migration yet. So, this is a bit of hand waving.

_storage/overview.md
moxiegirl: I've rewritten this section.

_storage/authentication.md
moxiegirl: The last one...it is a bit "spoken" structured. Corrected.

_storage/authentication.md
moxiegirl: cut and paste. Thanks for catching.

_storage/write-to-read.md
moxiegirl: blockstack.js in the Reference section.

_storage/write-to-read.md
moxiegirl: the hub_info is the write to url that you should see in the putFile examples

_storage/amazon-s3-deploy.md
moxiegirl: It says this because you went to the **EC2** dashboard rather than the AWS dashboard.

_storage/amazon-s3-deploy.md
moxiegirl: I provide an example to show them the result sans the angle brackets.  Most people understand the role of the angle brackets but those who don't can check character by character.  I'll drop in a bit about (be sure not to leave the brackets)

moxiegirl:
@kantai @markmhx Thank you for your review comments. I have addressed all of them and pushed my updates.  Please complete your final review.

cc: @wileyj

_storage/amazon-s3-deploy.md
markmhx: I mean to ask, should we just place the staging value as a hardcoded `0` value in the pastable blog instead of including it as a variable? It seems like something that only advanced users would want to change, and they'd know to spot and modify it themselves?

_storage/write-to-read.md
markmhx: In those examples, it appears that the `write-to` URL is "https://myhub.service.org/store/" but I'm not sure where that originates? 

That appears to contain the hub's origin (just judging by the "myhub" part) but also an additional path (`/store`)

_includes/architecture.md
markmhx: This is where the terminology draft sheet that I owe you would come in handy 😬

Here are the terms that mainly come to mind here for me:

- "Gaia hub": This is indeed a _service_ used by Blockstack users, and possibly set up by them as well. All hubs map one-to-many to storage providers (per below) and are necessary for users to connect to storage providers (i.e. Gaia doesn't connect users to storage providers directly). Gaia hubs can be hosted with the same company / general platform as storage providers (e.g. AWS with Gaia hub on EC2 and storage provider "on" / "as" S3) but don't have to be necessarily (e.g. Gaia hub on EC2, storage provider on DigitalOcean Spaces). We could call these "data hubs" as well, if we want a synonym that's more explicit. Though calling "Gaia hub URLs" also "data storage URLs" seems to risk confusion with "storage providers" below.
- "Storage provider": This is where the user's data "lives" having been written by users via Gaia hubs above (and made available to apps as reads via Gaia hubs as well). Users do not ever connect directly to storage providers – all interactions go through Gaia hubs, and no URLs should ever be set by users in the Blockstack UI for storage providers per se.

It appears we bungled the distinction between these two a bit in [the last modal shown for the new onboarding UI](https://forum.blockstack.org/t/user-choice-of-gaia-hub-now-available-during-onboarding/7026) in that we say "Appy recommends you store your data with this provider" (aka "storage provider") but then guide the user's attention to field for a _Gaia hub_ URL. 

We also generally spoke about how the user should choose or indicate a storage provider for usage (e.g. Blockstack's official provider), and yet to do so, we solicited Gaia hub URLs. Our language is technically pretty consistent (aside from the above), but this is easily confusing for users regardless.

Specifically to this part of the documentation, we seem to be lacking a good name for where Gaia hubs are hosted. I think you're recommending "hub provider", but since "provider" is a word already used for "storage provider", perhaps we should call these "Gaia hub hosts" or "data hub hosts"? I think saying that we run hubs (or "hub services" aka running instances of hubs, to be more specific) on "storage providers" is going to cause a lot of confusion since the name for data's "home" would be the same as the computing environment for a Gaia hub.

Side note – I do think many users are likely to care about the storage provider location and capabilities in addition to the Gaia hub's location and capabilities. In fact, they might care even more for the underlying storage provider than the Gaia hub, since the storage provider dictates several important qualities possibility (e.g. if my underlying storage provider is my Dropbox account, I know how to access it with a web UI outside of Blockstack and it gets synced to my desktop automatically).

That may be true even though lots of _other_ users don't care at all about their storage providers and perhaps more about the Gaia hubs (e.g. user wants to know they can ring the neck of a Gaia hub host they use at any time, should they need their data "out" of the system. Or they're mainly concerned about Gaia hub service uptime and latency).

_storage/hello-hub-choice.md
```diff
@@ -0,0 +1,262 @@
...
+      'http://localhost:8080/manifest.json',
+      ['store_write', 'publish_data'],
+      'http://localhost:8080/',
+      blockstack.nextHour().getTime(), {
+        solicitGaiaHubUrl: true
+      } // new options param
+    );
+    ```
+
+    {% include note.html content="If your app is running a different port than <code>8080</code>, enter that port instead. " %}
```
markmhx: Mine runs on 5000 with the default setup above – is that not the case for everyone?

_storage/hello-hub-choice.md
```diff
@@ -0,0 +1,262 @@
...
+
+   The extra `solicitGaiaHubUrl` parameter of `true` will cause the Blockstack Browser to prompt new identity creators for a storage hub URL.
+
+5. Save and close the `public/app.js` file.
+6. Make sure your app rebuilds cleanly.
```
markmhx: I made the above changes and ran `npm start` manually again, but I'm not encountering the new UI per testing below.

This is what my app.js shows: https://gist.github.com/markmhx/098132dc82ee2927a86e111d1fcd92b8

markmhx: Same (non-) result when I ensure the authRequest constant is within the event listener: https://gist.github.com/markmhx/fb97fadc08425e31aed8eb1ee39ade29

_storage/overview.md
kantai: Yes, the profile is usually stored in Gaia storage. But you're correct that it is somewhat of a special Gaia file. The blockchain explorer contains the exact URL of the profile file, whereas other Gaia files' URLs are obtained through reading information in the user's profile.

Because of this, if a user had a great command of the CLI, they _could_ store their profile file somewhere else, and still have it point to Gaia storage for their application storage, etc.

_storage/write-to-read.md
kantai: That's no longer the case in the newest images, @markmhx -- the write-to URL is "https://myhub.service.org"

_storage/amazon-s3-deploy.md
kantai: Everyone would want to change it to `1` I think

_storage/gaia-admin.md
```diff
@@ -0,0 +1,338 @@
...
+$ export API_KEY="hello"
+$ curl -H "Authorization: bearer $API_KEY" http://localhost:8009/v1/admin/config
+{"config":{"whitelist":["15hUKXg1URbQsmaEHKFV2vP9kCeCsT8gUu"]}}
+```
+
+To set the whitelist, you must set the *entire* whitelist. To set the list, pass a command such as the following:
+
+```bash
+$ export API_KEY="hello"
+$ curl -H "Authorization: bearer $API_KEY" -H 'Content-Type: application/json' -X POST --data-raw '["1KDcaHsYJqD7pwHtpDn6sujCVQCY2e1ktw", "15hUKXg1URbQsmaEHKFV2vP9kCeCsT8gUu"]' http://localhost:8009/v1/admin/config
```
kantai: This line should be:

```bash
$ export API_KEY="hello"
$ curl -H "Authorization: bearer $API_KEY" -H 'Content-Type: application/json' -X POST --data-raw '{"whitelist": ["1KDcaHsYJqD7pwHtpDn6sujCVQCY2e1ktw", "15hUKXg1URbQsmaEHKFV2vP9kCeCsT8gUu"]}' http://localhost:8009/v1/admin/config
```

_storage/gaia-admin.md
```diff
@@ -0,0 +1,338 @@
...
+
+This endpoint is used to read and write a Gaia hub's non-driver-related
+settings.  These include the port it listens on, and its proof-checking
+settings.
+
+To read the Gaia hub settings, you would run the following:
+
+```bash
+$ export API_KEY="hello"
+$ curl -H "Authorization: bearer $API_KEY" http://localhost:8009/v1/admin/config {"config":{"port":4000,"proofsConfig":{"proofsRequired":0}}}
```
kantai: I think this should just be:

```bash
$ export API_KEY="hello"
$ curl -H "Authorization: bearer $API_KEY" http://localhost:8009/v1/admin/config
```

(No need for the config options)

_storage/gaia-admin.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Using the admin service APIs
+
+You use the admin service APIs to manage the hub. Get calls require that you pass the `API_KEY` set in the admin service config:
+
+```bash
+$ export API_KEY="hello"
+```
+
+`POST` calls require that you supply your AWS access and secret key pair.
```
kantai: We don't need to set the AWS access and secret key pairs for these CURL requests.

_storage/gaia-admin.md
```diff
@@ -0,0 +1,338 @@
...
+         <td><code>setgid</code></td>
+         <td>This is the GID under which the command will run. This is optional.</td>
+      </tr>
+   </tbody>
+</table>
+
+
+## Using the admin service APIs
+
+You use the admin service APIs to manage the hub. Get calls require that you pass the `API_KEY` set in the admin service config:
```
kantai: We should add for people who set up their hub service via the AMI:

If you set up your hub service via the AMI, the admin interface is only accessible from within your virtual machine. To execute admin functions on your gaia hub, you'll need to ssh into your virtual machine:

```
ssh -t -i <your keyfile.pem> -A core@<public ip address>
```

_storage/hello-hub-choice.md
kantai: Yep, it ran on 5000 for me as well

_storage/hello-hub-choice.md
markmhx: Hm strange. I had reset my browser and gone through the create ID flow 🤔

_storage/amazon-s3-deploy.md
moxiegirl: Added a note this morning.

_includes/architecture.md
moxiegirl: A service such as Dropbox or My PC is uniquely different than an AWS / Digital Ocean. That type of service is offering storage at the individual consumer or small corporate level. While you could store your data there, **you can't run a hub service on a Dropbox, Apple Cloud, or My PC account**. 

https://www.dropboxforum.com/t5/Dropbox/Can-I-install-a-software-program-to-Dropbox/idi-p/41293

There is no operating system on which to run a hub service.  They do this for security reasons.  You can configure the Gaia hub service's configuration to store data on Dropbox et. all  --- but the Gaia URL points to the service, that service writes the data, and must live on a compute resource; AWS, Azure, or eventually a local machine. Caveat to local machine it would need to be up 24/7 and open to the internet. 

As for interacting with the data store outside of the hub service, why would you do that? It is the equivalent of editing a file by touching the bits on your disk. Possible in the future we might have applications that allow it.  There nearest analogy I could think of is moving my 1Password file around.  Sure, I can do that, and have, but I never view or modify the file's contents as that would no doubt cause 1Password to stop functioning.

_storage/overview.md
moxiegirl: @kantai Thank you.  At this point, I don't think this detail is necessary here. I do think it might be useful in the profile information. I've made a ticket to address this in the core documentation:

https://github.com/blockstack/docs.blockstack/issues/137

_storage/amazon-s3-deploy.md
moxiegirl: Did this.

_storage/hello-hub-choice.md
moxiegirl: Yeah, it kinda depends what port is available. I have a couple of services like a Jekyll build that takes the default port and I was running multiple hello-world. So, I mention that here your port may be different here so people who are have a similar situation can troubleshoot for themselves.

_storage/hello-hub-choice.md
moxiegirl: Resetting is covered in Step 4.

>     If you are already authenticated with an existing ID, you can choose that ID
   or **Deny**. Choosing an existing ID, signs you into the Hello Blockstack
   application, without offering a hub choice. Instead, choose **Deny**. This
   ends the authentication process. Then, reset the Blockstack Browser.

_storage/gaia-admin.md
moxiegirl: @kantai is that true even if I am making the call say from my local machine as opposed to from the instance terminal?

_storage/gaia-admin.md
moxiegirl: @wileyj and I had a conversation about this.  He said it was possible to make the curl calls from my local machine (after all just a curl).  If I do do that, I believe I need to set up local variables or use the AWS CLI to set up.  I've done this with similar environments in the past.

_storage/gaia-admin.md
kantai: Yep -- the AWS access and secret key pairs are only used if you are changing the driver backend to "aws" from "disk" (the change backend example later on in the doc uses those variables), so they should only be discussed in that section.

_storage/gaia-admin.md
kantai: Hmm -- I think you'd need to set up an ssh tunnel to make this work. As currently configured in the AMI, requests to `/admin` are blocked for everything but the localhost, and the port `8009` isn't configured to be open.

You could do something like this:

```
ssh -t -i <keyfile.pem> -L 8009:localhost:8009 -A core@<public ip address>
```

Which starts an SSH tunnel. Then you'd be able to use the curl commands from your local machine.

moxiegirl: fixed per meeting convo

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,478 @@
...
+
+   The system finds AMIs in the Marketplace and the Community.
+
+5. Choose **Community AMIs**.
+
+   The system displays the available Gaia hub images.
+
+   ![](/storage/images/gaia-community.png)
+
+6. Select the most recent version of the image.
```
markmhx: I'm finding it's hard to spot which AMI is the most recent by glancing – do I have to drill in somewhere to find that out?

screen shot 2019-02-20 at 15 24 17

markmhx: Or is the most recent one highest up in the list? If so, we should probably indicate that.

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,478 @@
...
+
+   At this point, your DNS management should look similar to the following except that with your domain rather than the `maryhub.ga` domain.
+
+   ![DNS fields](/storage/images/aws-dns-fields.png)
+
+7. After your changes propagate, visit your new domain at the `hub_info` page.
+
+   ![Domain test](/storage/images/domain-test.png)
+
+   If you receive another **Your connection is not private** dialogs, proceed to your domain.
```
markmhx: I'm not sure what "proceed to your domain" entails here exactly?

_storage/amazon-s3-deploy.md
```diff
@@ -0,0 +1,478 @@
...
+
+8. Refresh or renter the domain name fo your Gaia hub.
```
markmhx: Do we need a note here about how post-DNS-propagation, there may be an additional delay until the recurring background script that @wileyj created fully executes?

_storage/gaia-admin.md
moxiegirl: done.

_storage/gaia-admin.md
moxiegirl: removed those.

_storage/amazon-s3-deploy.md
moxiegirl: @markmhx Brought this up in the meeting yesterday. @wileyj is going to clean up the list; when he does, I'll retake the capture.  Then, there will be two images for each release. The images will have the same version but differ in the type of storage they provide.  The current version is 1.0.1 and user can choose ephemeral or ebs storage. I'll clarify this further in the docs.  I'll explain the version format in the docs. I think we can release with this capture and the explanation I just added.

_storage/amazon-s3-deploy.md
moxiegirl: The dialog literally has a proceed button this is a reference to it.

markmhx: Oh is that consistent across browser? If so, sounds good to me

_storage/amazon-s3-deploy.md
moxiegirl: done

PR #118, opened 2019-01-31
Fixes a newline and copypaste error in directory check

colindean:
Found while working through the tutorial while completing an interview project!


moxiegirl:
Thanks @colindean sorry this lingered so long. I definitely fell asleep on this one.

PR #124, opened 2019-02-07
[DO NOT MERGE] Democracy Earth Documentation

moxiegirl:
Adding content explaining how to vote with democracy earth


moxiegirl:
@paulamlb Here is the PR for the documentation.

PR #158, opened 2019-02-27
Improve app mining FAQ

friedger:
This PR
* adds a sentence about which apps are not eligable
* fixes a typo
* adds links to the details of the current reviewers and the app-mining github repo


moxiegirl:
@friedger Saw this and @GinaAbrams has some comments. So, we are not ignoring you.

PR #182, opened 2019-03-26
Add Android API Reference

friedger:
This PR
* adds a link to the (generated) Android Reference documentation


moxiegirl:
Thank @friedger I hadn't added this because the generated doc looks so...um awkward...course...pretty isn't required.

PR #245, opened 2019-05-10
Updates for Stacks Wallet v3

moxiegirl:
- Modifying emphasis > <em>
- Style updates to em and code
- Adding wallet security page
- Updated the glossary with security terms


moxiegirl:
> The example needs to be updated currently the wallet displays this message for fees and requires 0.00035076
> Screen Shot 2019-05-20 at 11 12 28 AM

BTW, my understanding is the fee for any transaction is market driven. None of my screencaps in terms of amounts will ever be an exact match. Users won't expect them to.  I believe that is what you were pointing out with this comment.

timstackblock:
Ok that makes sense if the number of BTC is not an issue we can proceed as is.

moxiegirl:
> Ok that makes sense if the number of BTC is not an issue we can proceed as is.
> 
> keyphrase.md doc has broken image icons, I know you said that you were still working on the images so this is probably already on your radar correct?

I am not seeing any project images.  What are you seeing?

markmhx:
@timstackblock @moxiegirl Are we looking good to release these changes by EOD today?

moxiegirl:
@markmhx I'm waiting for the installation DMGs to go live. Then I will push this live.

PR #263, opened 2019-05-24
Initial Smart Contract docs

moxiegirl:
Signed-off-by: Mary Anthony <mary@blockstack.com>


moxiegirl:
@kantai @jcnelson @lgalabru @zone117x The core draft is available for review. I believe I am caught up to what is merged in `blockstack/blockstack-core` as of today.  I regen'd the functions from the `clarity-preview-image`. 

I do have one section marked TODO in the reference on response types if Aaron or Jude can take a look there.  

My preferred method of receiving reviews is GitHub comments --- like any other code. Some folks find this uncomfy for docs. So, I will be emailing this content for review in PDF to you. You are welcome to comment on the PDF instead if that is preferable to you.

A preview build of this PR is produced by Netlify. You can also see this material on that preview and in context of the larger docs here.

https://deploy-preview-263--docsproduction.netlify.com/core/smart/overview.html

Please note, the section menu CSS is under dev. I'll have this working by the release date.

The SDK material is forthcoming. I have a quickstart from Ludo I'll work on and add to this material. I'll notify you when that is available.

moxiegirl:
image

_core/smart/overview.md
```diff
@@ -0,0 +1,135 @@
...
+
+Contracts can also call public functions from other smart contracts. The ability to do a static analysis of a smart contract allows a user to determine dependency between contracts.
+
+## The coding environment
+
+Clarity is a list processing (LISP) language, as such it is not compiled. Omitting compilation prevents the possibility of error or bugs introduced at the compiler level. You can write Clarity smart contract programs on any operating system with a text editor. You can use any editor you are comfortable with such as Atom, Vim, or even Notepad. The Clarity files you create with an editor have a `.clar` extension.
+
+Clarity is in pre-release and does not yet directly interact with the live Stacks blockchain. For the pre-release period you need a virtual machine (VM) environment to run Clarity contracts. Blockstack provides a Docker VM that you can use or you can build a test environment locally from code. Either the VM or a local environment is sufficient for testing Clarity programming for standalone contracts.
```
kantai: > you need a virtual machine (VM) environment

The use of "virtual machine" here could be a little confusing -- users just need to setup a local environment, not necessarily with a virtual machine or docker image.

moxiegirl: I want to say "test environment" which is either the VM or a build in their local environment. Does that work?

_core/smart/overview.md
```diff
@@ -0,0 +1,135 @@
...
+
+This program defines a single `hello-world` user function. The `begin` is a native Clarity function that evaluates the expressions input to it and returns the value of the last expression.  Here there is a single `print` expression. Both the `begin` and the `print` are enclosed in `()` parentheses.
+
+For the pre-release, the Blockstack virtual machine (VM) includes the `clarity-cli` command for interacting with the contract and SQLite to support the data space. You create a SQLLite database to hold data related to Clarity contracts. This database simulates the blockchain by recording the contract activity. 
+
+You can't run even an a hello-world program without first initializing a Clarity contract's data space within the database. You can use the `clarity-cli initialize` command to set up the database.
+
+```clarity-cli initialize /data/db```
+
+This command records the contract in the `db` database which resides in the `/data` directory of the VM.
```
kantai: This command doesn't record a contract yet -- it just initializes the db.

_core/smart/overview.md
```diff
@@ -0,0 +1,135 @@
...
+contracts            maps_table           type_analysis_table
+data_table           simmed_block_table 
+sqlite> 
+```
+
+After you initialize the contract's data space, you can `check` a Clarity program for problems.
+
+```clarity-cli check ./hello.clar /data/db```
+
+As the name implies, the `check` ensures the contract definition passes a type check; passing programs return. Once a contract passes a check, you `launch` it.
```
kantai: > passing programs return.

I think this should be "passing programs will return an exit code of 0"

_core/smart/principals.md
```diff
@@ -0,0 +1,67 @@
...
+
+A smart contract may use the special variable `contract-principal` to refer to its own principal.
+
+To allow smart contracts to operate on assets it owns, smart contracts may use the special `(as-contract expr)` function. This function executes the expression (passed as an argument) with the `tx-sender` set to the contract's principal, rather than the current sender. The `as-contract` function returns the value of the provided expression.
+
+For example, a smart contract that implements something like a "token faucet" could be implemented as so:
+
+```cl
+(define-public (claim-from-faucet)
+  (if (is-none? (fetch-entry claimed-before (tuple (sender tx-sender))))
```
kantai: Eventually, we'll probably add support for multi-line `let` definitions, but for now, something like this needs to be wrapped in a begin:

```
(let ((requester tx-sender)) ;; set a local variable requester = tx-sender
    (begin
        (insert-entry! claimed-before (tuple (sender requester)) (tuple (claimed 'true)))
        (as-contract (stacks-transfer! requester 1)))))
```

_core/smart/functions.md
```diff
@@ -0,0 +1,151 @@
...
+{:.no_toc}
+
+Clarity includes _defines_  and native functions for creating user-defined functions. 
+
+* TOC
+{:toc}
+
+## define and define-public functions
+
+Functions specified via `define-public` statements are public functions. Functions without these designations, simple `define` statements, are private functions. You can run a contract's public functions directly via the `clarity-cli execute` command line directly or from other contracts. You can use the `clarity eval` or `clarity eval_raw` commands to evalut private functions via the command line.
```
kantai: Typo: "evalut" should be "evaluate"

_core/smart/functions.md
```diff
@@ -0,0 +1,151 @@
...
+
+Public functions return a Response type result. If the function returns an `ok` type, then the function call is considered valid, and any changes made to the blockchain state will be materialized. If the function returns an `err` type, it is considered invalid, and has no effect on the smart contract's state. 
+
+For example, consider two functions, `foo.A` and `bar.B` where the `foo.A` function calls `foo.B`, the table below shows the data materialization that results from the possible combination of return values:
```
kantai: I think `foo.B` should be `bar.B`

_core/smart/functions.md
```diff
@@ -0,0 +1,151 @@
+---
+layout: core
+permalink: /:collection/:path.html
+---
+# Using functions
```
kantai: This section describes more than functions -- it seems like "Defining Functions and Data Maps" would fit more closely.

_core/smart/functions.md
```diff
@@ -0,0 +1,151 @@
...
+
+For intra-contract calls dynamic dispatch is not supported. When a contract is launched, any contracts it depends on (calls) must exist. Additionally, no cycles may exist in the call graph of a smart contract. This prevents recursion (and re-entrancy bugs. A static analysis of the call graph detects such structures and they are rejected by the network.
+
+A smart contract may not modify other smart contracts' data directly; it can read data stored in those smart contracts' maps. This read ability does not alter any confidentiality guarantees of Clarity. All data in a smart contract is inherently public, andis readable through querying the underlying database in any case.
+
+To read another contract's data, use `(fetch-contract-entry)` function. This behaves identically to `(fetch-entry)`, though it accepts a contract principal as an argument in addition to the map name:
+
+```cl
+(fetch-contract-entry
+  'contract-principal
```
kantai: For now, it makes more sense for this to be `contract-name` rather than `contract-principle`, because at the moment, names are used rather than the contract principal.

_core/smart/functions.md
```diff
@@ -0,0 +1,151 @@
...
+  'map-name
+  'key-tuple) ;; value tuple or none
+```
+
+For example, you could do this:
+
+```cl
+(fetch-contract-entry
```
kantai: This example is probably better  as:

```cl
(fetch-contract-entry
  names
  name-map
  1)
```

Note `name-map` doesn't require a single quote.

_core/smart/functions.md
```diff
@@ -0,0 +1,151 @@
...
+ 'SC3H92H297DX3YDPFHZGH90G8Z4NPH4VE8E83YWAQ
+ 'name-map
+ 1)     ;;returns owner principal of name represent by integer 12234
+```
+
+Just as with the `(contract-call)` function, the map name and contract principal arguments must be constants, specified at the time of publishing.
+
+Finally, and importantly, the `tx-sender` variable does not change during inter-contract calls. This means that if a transaction invokes a function in a given smart contract, that function is able to make calls into other smart contracts on your behalf. This enables a wide variety of applications, but it comes with some dangers for users of smart contracts. However, the static analysis guarantees of Clarity allow clients to know a priori which functions a given smart contract will ever call. Good clients should always warn users about any potential side effects of a given transaction.
+
+### Reading from Other Smart Contracts
```
kantai: This subsection looks like a repeat of the above.

_core/smart/tutorial.md
```diff
@@ -0,0 +1,335 @@
...
+
+    ```bash
+    root@c28600552694:/src/blockstack-core# cat sample-programs/tokens.clar 
+    ```
+
+    The next section gives you an introduction to the Clarity language by way of examining this program's code.
+
+## Task 2: Review a simple Clarity program
+
+If you haven't already done so, use the `cat` or `more` command to dipslay the `tokens.clar` file's code. Clarity is designed for static analysis; it is not a compiled language and is not Turing complete. It language is a LISP-like language.  LISP is an acronym for list processing. the first line of the program contains a user-defined `get-balance` function.
```
kantai: Typo: 'dipslay'

_core/smart/tutorial.md
```diff
@@ -0,0 +1,335 @@
...
+If you haven't already done so, use the `cat` or `more` command to dipslay the `tokens.clar` file's code. Clarity is designed for static analysis; it is not a compiled language and is not Turing complete. It language is a LISP-like language.  LISP is an acronym for list processing. the first line of the program contains a user-defined `get-balance` function.  
+
+```cl
+(define (get-balance (account principal))
+  (default-to 0 (get balance (fetch-entry tokens (tuple (account account))))))
+```
+
+`get-balance` is a private function because it is constructed with the `define` call. To create public functions, you would use the `define-public` function. Public functions can be called from other contracts or even from the command line with the `clarity-cli`.
+
+Notice the program is enclosed in  `()` (parentheses) and each statement as well.  The `get-balance` function defines an `account` field which takes the special type `principal`, Principals represent a spending entity and are roughly equivalent to a Stacks address.
```
kantai: It'd be more accurate to say "The get-balance function takes an `account` argument of the special type `principal`,..."

_core/smart/tutorial.md
kantai: I think it'd be helpful to say here that we're currently looking at the `tokens.clar` sample.

_core/smart/tutorial.md
```diff
@@ -0,0 +1,335 @@
...
+(define (get-balance (account principal))
+  (default-to 0 (get balance (fetch-entry tokens (tuple (account account))))))
+```
+
+`get-balance` is a private function because it is constructed with the `define` call. To create public functions, you would use the `define-public` function. Public functions can be called from other contracts or even from the command line with the `clarity-cli`.
+
+Notice the program is enclosed in  `()` (parentheses) and each statement as well.  The `get-balance` function defines an `account` field which takes the special type `principal`, Principals represent a spending entity and are roughly equivalent to a Stacks address. 
+
+Along with the `principal` types, Clarity supports  booleans, integers, and fixed length buffers. Variables are created via `let` binding but there is no support for mutating functions like `set`.
+The `names.clar` code includes two public functions.
```
kantai: > The `names.clar` code includes two public functions.  
This seems out of place --I think we're still looking at `tokens.clar` at this point.

_core/smart/tutorial.md
```diff
@@ -0,0 +1,335 @@
...
+  (if (<= tokens 0)
+      (err "must move positive balance")
+      (let ((current-amount (get-balance account)))
+        (begin
+          (set-entry! tokens (tuple (account account))
+                      (tuple (balance (+ tokens current-amount))))
+          (ok tokens)))))
+```
+
+Every smart contract has both a data space and code. The data space of a contract may only interact with that contract. This particular contract is named `tokens`
```
kantai: Should read: "This particular function is interacting with a map named `tokens`"

_core/smart/tutorial.md
```diff
@@ -0,0 +1,335 @@
...
+      (err "must move positive balance")
+      (let ((current-amount (get-balance account)))
+        (begin
+          (set-entry! tokens (tuple (account account))
+                      (tuple (balance (+ tokens current-amount))))
+          (ok tokens)))))
+```
+
+Every smart contract has both a data space and code. The data space of a contract may only interact with that contract. This particular contract is named `tokens` 
+The `set-entry!` function is a native function that sets the value associated with the input key to the inputted value in the `tokens` data space. Because `set-entry!`  mutates data so it has an `!` exclamation point; this is by convention in Clarity.
```
kantai: `tokens` is more accurately described as a data-map than a data space.

_core/smart/tutorial.md
```diff
@@ -0,0 +1,335 @@
...
+
+(token-credit! 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR 10000)
+(token-credit! 'SM2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQVX8X0G 300)
+```
+
+The final two lines of the program pass a principal, represented by a Stacks address, and an amount to the private user-defined `token-credit` function.
+
+Smart contracts may call other smart contracts using a `contract-call!` function. This means that if a transaction invokes a function in a given smart contract, that function is able to make calls into other smart contracts on your behalf. The ability to read and do a static analysis of Clarity code allows clients to learn which functions a given smart contract will ever call. Good clients should always warn users about any potential side effects of a given transaction.
+
+Take a moment to `cat` the contents of the `name.clar` file.
```
kantai: Typo: `name.clar` -> `names.clar`

_core/smart/tutorial.md
```diff
@@ -0,0 +1,335 @@
...
+
+    This happens because the `names.clar` contract _calls_ the `tokens.clar` contract, and that contract has not been created on the blockchain.
+
+3. Type check the `tokens.clar` contract, it should pass a check as it does not use the `contract-call` function:
+
+    ```bash
+    # clarity-cli check sample-programs/tokens.clar /data/db
+    ```
+
+    When the `check` command executes successfully, it does not output any information, it just exits with the stand UNIX `0` exit code.
```
kantai: This now outputs "Checks passed.":

```
# clarity-cli check sample-programs/tokens.clar /data/db
Checks passed.
```

_core/smart/tutorial.md
```diff
@@ -0,0 +1,335 @@
...
+4. Launch the `tokens.clar` contract.
+
+   You use the `launch` command to instantiate a contract on the Stacks blockchain. If you have dependencies between contracts, for example names.clar is dependent on tokens.clar, you must launch the dependency first.
+
+    ```bash
+    # clarity-cli launch tokens sample-programs/tokens.clar /data/db
+    Contract initialized!
+    ```
+    
+    Once launched, you can execute the contract or a public method on the contract. Your development database has an instantiated `tokens` contract. If you were to close the VM, you and relaunch it later with the same mount point and you wouldn't need to relaunch.
```
kantai: > If you were to close the VM, you and relaunch it later with the same mount point and you wouldn't need to relaunch.

Could be:

If you were to close the VM and relaunch it later with the same mount point, you wouldn't need to relaunch the contract-- the database has persisted the tokens contract.

_core/smart/overview.md
kantai: Yep that works

_core/smart/clarityRef.md
```diff
@@ -0,0 +1,193 @@
...
+
+This section lists the types available to smart contracts. The only atomic types supported by the Clarity are booleans, integers, fixed length buffers, and principals. 
+
+### Int type
+
+The integer type in the Clarity language is a 16-byte signed integer, which allows it to specify the maximum amount of microstacks spendable in a single Stacks transfer. The special `BlockHeightInt` you can obtain with the `get-block-info` function.
+
+### Bool type
+
+Public functions return a boolean result of `'true` or `'false`. If the function returns `'true`, then it is considered valid, and any changes made to the blockchain state are materialized. If the function returns `'false`, the function is considered invalid and does not effect the smart contract's state.
```
kantai: This is no longer the case -- public functions now return "Response" types (it's described correctly in the functions.html file)

_core/smart/clarityRef.md
```diff
@@ -0,0 +1,193 @@
...
+
+Clarity supports lists of the atomic types. However, the only variable length lists in the language appear as function inputs; there is no support for list operations like append or join.
+
+### Principal type
+
+Clarity provides this primitive for checking whether or not the smart contract transaction was signed by a particular principal. Principals represent a spending entity and are roughly equivalent to a Stacks address. The principal's signature is not checked by the smart contract, but by the virtual machine. A smart contract function can use the  globally defined `tx-sender` variable to obtain the current principal.
+
+Smart contracts may also be principals (represented by the smart contract's identifier). However, there is no private key associated with the smart contract, and it cannot broadcast a signed transaction on the blockchain. A smart contract uses the special variable `contract-principal` to refer to its own principal.
+
+You can use the `is-contract?` to determine whether a given principal corresponds to a smart contract.
```
kantai: This function isn't defined yet, but will be eventually.

_core/smart/clarityRef.md
```diff
@@ -0,0 +1,193 @@
...
+
+### String type
```
kantai: There's no string type, all strings are Buffers

_core/smart/clarityRef.md
```diff
@@ -0,0 +1,193 @@
...
+```cl
+(hash160 "aaron.id")
+```
+
+### Tuple type
+
+To support the use of named fields in keys and values, Clarity  allows the construction of named tuples using a function `(tuple ...)`, for example
+
+```cl
+(define imaginary-number-a (tuple #real 1 #i 2))
```
kantai: The syntax for tuples no longer uses `#`s, so this example would be:

```
(define imaginary-number-a (tuple (real 1) (i 2)))
(define imaginary-number-b (tuple (real 2) (i 3)))
```

_core/smart/clarityRef.md
```diff
@@ -0,0 +1,193 @@
...
+This allows for creating named tuples on the fly, which is useful for data maps where the keys and values are themselves named tuples. Values in a given mapping are set or fetched using:
+
+<table class="uk-table uk-table-small">
+<tr>
+  <th class="uk-width-small">Function</th>
+  <th>Description</th>
+</tr>
+<tr>
+  <td><code>(fetch-entry map-name key-tuple)</code></td>
+  <td>Fetches the value associated with a given key in the map, or returns <code>'null</code> if there is no such value.</td>
```
kantai: Instead of `'null`, it'd be `none`

_core/smart/clarityRef.md
```diff
@@ -0,0 +1,193 @@
...
+  <td>Sets the value of key-tuple in the data map if and only if an entry does not already exist.</td>
+</tr>
+  <tr>
+   <td><code>(delete-entry! map-name key-tuple)</code></td>
+  <td>Deletes key-tuple from the data map.</td>
+</tr>
+</table>
+
+
+To access a named value of a given tuple, the `(get #name tuple)` function returns that item from the tuple.
```
kantai: There's no more use `#` in `(get ...)`.

Now it'd be `(get name tuple)`.

_core/smart/clarityRef.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Optional type
+
+Represents an optional value. Use the optional type for:
```
kantai: This might be more clear:

"Represents an optional value. This is used in place of the typical usage of "null" values in other languages, and represents a type that can either be _some value_ or _none_". These types are used as the return types of data-map functions.

_core/smart/clarityRef.md
```diff
@@ -0,0 +1,193 @@
...
+
+* initial values
+* return values for functions
+* return value for otherwise reporting simple errors
+
+If a `get` function is passed an `Optional&gt;Tuple&lt;`
+
+### Response type
```
kantai: Response types represent the result of a public function, and are used to indicate and return data associated with the execution of the function, as well as whether the function error'ed (and therefore did not materialize any data in the database) or ran ok (in which case data materialized in the database).

Response types contain two subtypes -- a response type in the event of `ok` (e.g., a public function returns an integer code on success) and an `err` type (e.g., a function returns a buffer on error).

_core/smart/clarityRef.md
moxiegirl: Is it correct to say:

Supports values of `'true` or `'false`.

kantai: Yep

_core/smart/clarityRef.md
moxiegirl: commented out.

_core/smart/clarityRef.md
moxiegirl: Removed this section. Can you check the revised Buffet type section?

_core/smart/clarityRef.md
moxiegirl: Fixed.

_core/smart/clarityRef.md
moxiegirl: 🆗

_core/smart/clarityRef.md
moxiegirl: 🙏

_core/smart/clarityRef.md
moxiegirl: 🙏

_core/smart/functions.md
moxiegirl: ✔️

_core/smart/functions.md
moxiegirl: ✔️

_core/smart/functions.md
moxiegirl: ✔️ @kantai  this doesn't require a source change in the repo does it?

_core/smart/functions.md
moxiegirl: yep.

_core/smart/overview.md
moxiegirl: Cool. Changed through out the doc.

_core/smart/overview.md
moxiegirl: Fixed

_core/smart/overview.md
moxiegirl: Thanks.

_core/smart/principals.md
moxiegirl: ✔️

_core/smart/functions.md
kantai: I don't think so

_core/smart/tutorial.md
moxiegirl: Yep, I was aware of the duplication but some people just jump to the tutorial.

_core/smart/tutorial.md
moxiegirl: ✔️

_core/smart/tutorial.md
moxiegirl: fixed

_core/smart/tutorial.md
moxiegirl: ✔️

_core/smart/tutorial.md
moxiegirl: check.

_core/smart/tutorial.md
moxiegirl: check

_core/smart/tutorial.md
moxiegirl: check.

_core/smart/tutorial.md
moxiegirl: check

_core/smart/clarityRef.md
kantai: Yep ---

I'd revise the current Buffer section as:

```
Buffer types represent fixed-length byte buffers. Currently, the only way to construct a Buffer is using string literals, for example:

`"alice.id"`

or

`hash160("bob.id")`

All of the hash functions return buffers:

hash160
sha256
keccak256

The block properties `time`, `header-hash`, `burnchain-header-hash`, and `vrf-seed` are also all buffers.
```

_core/smart/tutorial.md
kantai: Okay, works for me

_core/smart/clarityRef.md
zone117x: I think block property `time` is an int

zone117x: > Currently, the only way to construct a Buffer is using string literals

What about hex literals? `0xdeadbeef` ?

_core/smart/functions.md
moxiegirl: fixed

_core/smart/functions.md
moxiegirl: Done.

_core/smart/clarityRef.md
moxiegirl: @kantai I have no answer for @zone117x --- do you have a response?

kantai: Yes -- hex literals would work as well.

PR #265, opened 2019-05-27
fixed typo

danosaur98:


moxiegirl:
Nice catch! Thank you @danosaur98

PR #269, opened 2019-06-04
Update blockstack_storage.md

hongily25:
Got an uncaught ReferenceError: scopes is not defined.

image


moxiegirl:
@hongily25 Thank you for the catch!

PR #274, opened 2019-06-06
Update zero_to_dapp_2.md

danosaur98:
fixed typo


moxiegirl:
@danosaur98 Thank you for the catch and the PR!

PR #282, opened 2019-06-21
Deleted steps in storage tutorial made redundant by new React App Generator

Sahana-Srinivasan:
For example, the setting to allow dots in the url is enabled by default by create-react-app, and the BrowserRouter component is already in the src/index.js file.  (Only relevant after the React app generator changes have been approved)


moxiegirl:
Pull this after the https://github.com/blockstack/blockstack-app-generator/pull/54 is merged.

PR #377, opened 2019-09-16
Collections preview documentation

moxiegirl:


_develop/collections.md
```diff
@@ -0,0 +1,217 @@
...
+
+If you have `npm` installed, do the following to run the Contact Manager demo app:
+
+1. If you have a local Blockstack installed, <a href="{{"/browser/local_browser.html" | prepend:site.baseurl}}" >uninstall it</a>.
+
+2. Download and install the <a href="https://github.com/blockstack/blockstack-browser/releases/tag/collections-alpha.1" target="_blank">Collections Alpha Build</a> of the Blockstack Browser client for your OS.
+
+3. Launch the alpha build of the local Blockstack Browser client.
+
+   {% include warning.html content="This alpha build interacts with a version of the Gaia Storage hub that supports collections. You cannot use an existing ID to test with, instead, you need to <strong>create a new, test ID using the alpha build</strong>. Additionally, <strong>you must create a custom Gaia hub during the Contact Manager onboarding</strong>. For this reason, wait to create a new ID." %}
```
yknl: They don't necessarily need to create a custom Gaia hub. They just need to connect to the custom one that we put up for the preview.

_develop/collections.md
```diff
@@ -0,0 +1,217 @@
...
+   "firstName": "sally",
+   "blockstackID": "",
+   "email": "",
+   "website": "",
+   "telephone": "",
+   "identifier": "sally jeffries"
+}
+```
+
+A collection schema is neither validated or enforced. The goal is to incentivize collection use rather that enforce use. Developers have the ability to roll-back in case apps make undesirable changes that break compatibility with collections. To support this rollback in the pre-release, collections data store is conceptually an event log. Every data write an app makes is stored as a separate file. By placing data in files it ensures that data is never lost and developers can return files back to any previous state.
```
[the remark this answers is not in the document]
moxiegirl: @yknl Thanks, that was unclear in the original. Can users do this in the preview? Or is this something we will enable in the future?

_develop/collections.md
```diff
@@ -0,0 +1,217 @@
+---
+layout: learn
+permalink: /:collection/:path.html
+---
+# Work with Collections (Preview)
+{:.no_toc}
+
+Collections is that feature designed to make data portable among Blockstack applications. Sharing is accomplished by storing a user's data in a standard, structured way in a known Gaia location. Collections associate the data with a user's distributed ID. When users move among apps, the same data is available to each application.
```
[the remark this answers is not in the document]
moxiegirl: @yknl yep, looks like a typo. I'll fix that.

_develop/collections.md
```diff
@@ -0,0 +1,217 @@
...
+   "telephone": "",
+   "identifier": "sally jeffries"
+}
+```
+
+A collection schema is neither validated or enforced. The goal is to incentivize collection use rather that enforce use. Developers have the ability to roll-back in case apps make undesirable changes that break compatibility with collections. To support this rollback in the pre-release, collections data store is conceptually an event log. Every data write an app makes is stored as a separate file. By placing data in files it ensures that data is never lost and developers can return files back to any previous state. 
+
+<div class="uk-card uk-card-default uk-card-body">
+<h5 class="uk-card-title">The Future of Collections Envisioned</h5>
+<p>Blockstack believes that collections feature of the Blockstack platform can reduce the damage that faulty and malicious apps can cause to user’s data. For end=users, Blockstack seeks to enable:</p>
```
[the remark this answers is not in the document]
moxiegirl: Thanks, I'll clarify.

_develop/collections.md
moxiegirl: Ok, I'll clarify.

_develop/collections.md
```diff
@@ -0,0 +1,215 @@
...
+    import { UserSession, AppConfig, makeAuthRequest } from 'blockstack'
+    import { Contact } from '`blockstack-collections'
+
+    const scopes = ['store_write', 'publish_data', Contact.scope]
+    const appConfig = new AppConfig(scopes)
+    const userSession = new UserSession({appConfig: appConfig})
+
+    const authRequest = makeAuthRequest(undefined, undefined, undefined, scopes, undefined, undefined, {
+    solicitGaiaHubUrl: true,
+    recommendedGaiaHubUrl: 'https://develop-hub.blockstack.org'
```
zone117x: This should be `https://staging-hub.blockstack.xyz`

zone117x:
A few areas are using `https://develop-hub.blockstack.org` but should be using `https://staging-hub.blockstack.xyz`

_develop/collections.md
```diff
@@ -0,0 +1,215 @@
+---
+layout: learn
+permalink: /:collection/:path.html
+---
+# Work with Collections (Preview)
+{:.no_toc}
+
+Collections is the feature designed to make data portable among Blockstack applications. Sharing is accomplished by storing a user's data in a standard, structured way in a known Gaia location. Collections associate the data with a user's distributed ID. When users move among apps, the same data is available to each application.
```
yknl: Does this sound better?
"Sharing is accomplished by storing a user's data in a standardized format in a known location."

Also I don't think we've used the term "distributed ID" before.

_develop/collections.md
```diff
@@ -0,0 +1,215 @@
...
+
+{% include note.html content="This is a preview release of the <code>Contact</code> collections is a preview release. This release allows developers to try out the new collections functionality and we are interested in collecting feedback. Please feel free to report issues or request enhancements with collections or <code>Contacts</code> themselves on the <a href='https://github.com/blockstack/blockstack-collections/issues/new' target='_blank'>blockstack/blockstack-collections</a> repository.  If you encounter problems with <code>blockstack.js</code> you can <a href='https://github.com/blockstack/blockstack.js/issues/new' target='_blank'>file issues or request enhancements on its repo</a>." %}
+
+## Understand how collections work
+
+One of Blockstack's goals is to give users true data ownership by enabling *data portability*.  Data portability allows users to login with their digital ID on any app and have access to the same data. For example, if a user adds a photo of a Hawaiian vacation in one app, that photo enters the user's data pool. Then, when the user opens a second app, that same photo is already in use because the user's digital ID login is identical for both apps.
```
yknl: We might want to consistently use decentralized ID instead of digital ID. All online IDs are digital IDs.

Consider changing the last part to:
"... that same photo will be available for use because data is linked to the same decentralized ID."

_develop/collections.md
moxiegirl: I don't think that change is significant either way. I do think it is important to mention Gaia.  I'll switch distributed ID for decentralized throughout.

_develop/collections.md
moxiegirl: Consistency in place. I restructured the sentence for clarity.

PR #383, opened 2019-09-24
git suggestions for the Blockstack Tutorial

nickycutesc:
Just some simple git problems I ran into and their solutions when going through the tutorial.

### DIFF
--- a/_browser/hello-blockstack.md
+++ b/_browser/hello-blockstack.md
@@ -353,12 +353,28 @@ Complete the tutorial by storing your app code on GitHub. Before you begin, make
     ```bash
     git remote add origin git@github.com:YOUR_USERNAME_HERE/hello-blockstack.git
     ```
+    
+5. If you see this error ('refusing to merge unrelated histories’), please run the following command:
 
-5. Push your new code to the master branch of the remote repo:
+    ```bash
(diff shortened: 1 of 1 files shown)


moxiegirl:
@nickycutesc Thanks for the PR.  Doing a quick review.

moxiegirl:
@nickycutesc The git information in this file is pretty old and assumes quite a bit. So, your changes even are environment-specific. I'm going to rewrite the entire section to simply suggest creating a git repo is a possibility and leave it at that.

Thank you for creating this PR and point out the problems.

nickycutesc:
Thanks @moxiegirl for reviewing my PR! Could I collaborate with you to rewrite the git section?

moxiegirl:
@nickycutesc Git is its own product and its use varies significantly depending on the user's OS and how it is set up (some users use a command line and others use one of many clients), maintaining doc in our product about how to use Git is not appropriate. It is better to refer people to Git documentation which will be both more extensive and formerly maintained.

 Happy to collaborate on something specific to Blockstack tech if you have suggestions there.

nickycutesc:
I’d have to look at the project some more to think of any ideas. What are some of the most pressing issues affecting Blockstack thus far?

PR #437, opened 2019-11-04
Update Clarity Supported Types with latest changes

lgalabru:


moxiegirl:
@lgalabru You did not mention the unsigned int in particular under the `Int` type. Was that intentional?

PR #454, opened 2019-11-15
Add instruction to readme

danigrant:
There was no description of how to get the bundle command so I added one

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ To run locally:
     ./get-content.sh
     ```
 
-3. Build and serve locally.
+3. Build and serve locally. You will need [bundler](https://bundler.io/). You can install it with `gem install bundler` and then can run `bundle install` before running this bundle exec command below:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,7 +13,7 @@ To run locally:
     ./get-content.sh
     ```
 
-3. Build and serve locally.
+3. Build and serve locally. You will need [bundler](https://bundler.io/). You can install it with `gem install bundler` and then can run `bundle install` before running this bundle exec command below:
```
moxiegirl: Hi @danigrant, thank you for the PR>

These instructions assume the user has already installed Jekyll properly.  So, this sub-section implies that isn't the case.  Would you like to instead like to add an intro sentence pointing users to the the Jekyll install if they haven't done that bit?

moxiegirl:
Closing. Please reopen if you have another approach.

PR #462, opened 2019-11-28
Clarify app mining scoring

njordhov:
Copyedit as some sentences were misplaced or unclear.


moxiegirl:
@GinaAbrams This language is verbatim as provided. Can you do a check and let me know if you agree with @njordhov 's changes.

PR #498, opened 2020-01-17
Add details about how to do reads

### LINKED ISSUE #344: Missing question from the technical FAQ
This question from the old technical FAQ comes up a lot, but appears to be missing from docs.blockstack.org:

```
## How can I read my public app data without `blockstack.js`?		
￼
￼ The URLs to a user's public app data are in a canonical location in their		
￼ profile.  For example, here's how you would get public data from the		
￼ [Publik](https://publik.ykliao.com) app, stored under the Blockstack ID `ryan.id`.		
￼
￼ 1. Get the bucket URL		
￼ ```bash		
￼ $ BUCKET_URL="$(curl -sL https://core.blockstack.org/v1/users/ryan.id | jq -r '."ryan.id"["profile"]["apps"]["http://publik.ykliao.com"]')"		
￼ $ echo "$BUCKET_URL"		
￼ https://gaia.blockstack.org/hub/1FrZTGQ8DM9TMPfGXtXMUvt2NNebLiSzad/		
￼ ```		
￼
￼ 2. Get the data		
￼ ```bash		
￼ $ curl -sL "${BUCKET_URL%%/}/statuses.json"		
￼ [{"id":0,"text":"Hello, Blockstack!","created_at":1515786983492}]		
￼ ```
```

Can't believe I missed it when we did the transition :sweat_smile:.  Please add it to the new docs.blockstack.org page?

friedger:
This PR
* adds a paragraph to `Address-based access-control` section about how to read (public) data.
* fixes #344 (using banter.pub here instead of publik due to server certificate problems)

### DIFF
--- a/_storage/write-to-read.md
+++ b/_storage/write-to-read.md
@@ -63,3 +63,16 @@ demonstrate that they control _that_ address. This is achieved via the
 authentication token which is a message _signed_ by the private key associated
 with that address. The message itself is a challenge text, returned via the
 `/hub_info/` endpoint.
+
+Reads can be done by everybody. The URLs to a user's app data are in a canonical location in their profile. For example, here's how you would get data from the
+￼ [Banter](https://banter.pub/) app, stored under the Blockstack ID `gavin.id`.
+￼
+￼ 1. Get the bucket URL
+￼ `bash ￼ $ BUCKET_URL="$(curl -sL https://core.blockstack.org/v1/users/gavin.id | jq -r '."gavin.id"["profile"]["apps"]["https://banter.pub"]')" ￼ $ echo "$BUCKET_URL" ￼ https://gaia.blockstack.org/hub/16E485MVpR3QpmjVkRgej7ya2Vnzu3jyTR/ ￼`
+￼
+￼ 2. Get the data
+￼ `bash ￼ $ curl -sL "${BUCKET_URL%%/}/Message/3e866af471d0-4072-beba-06ad1e7ad4bd" ￼ {"content":"Anyone here?","votes":[],"createdBy":"gavin.id",...} ￼`
+
+This data is public and unencrypted. The same works for encrypted data.
+Only the holder of the private key used for
+encryption would be able to decrypt the data.


_storage/write-to-read.md
```diff
@@ -63,3 +63,16 @@ demonstrate that they control _that_ address. This is achieved via the
 authentication token which is a message _signed_ by the private key associated
 with that address. The message itself is a challenge text, returned via the
 `/hub_info/` endpoint.
+
+Reads can be done by everybody. The URLs to a user's app data are in a canonical location in their profile. For example, here's how you would get data from the
```
moxiegirl: What do you mean by everybody?  Do you mean non-developers such as end-users? If end user, why would they want to do this? Most of our app users are not bash users... I'm trying to see the use case you are documenting.

friedger: Everybody in contrast to the owner of the private keys.

The use case is
* for developers to verify what data was written and to understand the difference between read and write.

friedger: * for end users to see the data they own.

Not sure whether that makes sense for the normal end-user. Maybe we can just show two screenshots of banter message as shown by the app and as stored on gaia.

moxiegirl: @friedger thanks for clarifying the use case. I think it makes sense for developers to know curl and use it.  Let me review it again with that in mind.

_storage/write-to-read.md
```diff
@@ -63,3 +63,16 @@ demonstrate that they control _that_ address. This is achieved via the
...
+￼ [Banter](https://banter.pub/) app, stored under the Blockstack ID `gavin.id`.
+￼
+￼ 1. Get the bucket URL
+￼ `bash ￼ $ BUCKET_URL="$(curl -sL https://core.blockstack.org/v1/users/gavin.id | jq -r '."gavin.id"["profile"]["apps"]["https://banter.pub"]')" ￼ $ echo "$BUCKET_URL" ￼ https://gaia.blockstack.org/hub/16E485MVpR3QpmjVkRgej7ya2Vnzu3jyTR/ ￼`
```
zone117x: I think this is a useful example and provides good context. Only issue is that `jq` is probably not installed on most machines. 
@friedger thoughts?

PR #513, opened 2020-02-10
Update mining_enroll.md

GinaAbrams:
Per today's community call and announcement.


GinaAbrams:
@hstove for your review. Think we keep the meaningful parts of app mining docs and remove the super specific sections on things like enrollment.

PR #539, opened 2020-04-10
Add note on outdated docs pages

agraebe:
ezgif-5-9785275a6be5

If a specific docs page hasn't been updated for 6 months, we add the following note:

> This page has been published more than 6 months ago, and parts of its contents are very likely to be obsolete by now.

WDYT @markmhx @jeffdomke @hstove?

A few notes:
- Text can easily be changed, just a draft
- This seems to show up on quite a few pages (randomly going through some)
- It can be a nice forcing function to identify content to either review or delete


_includes/post-meta.html
```diff
@@ -1,19 +1,31 @@
...
+<div>
+  {% assign three_months_ago = 'now' | date: '%s' | minus: 15552000 %}
+  {% assign page_date = page.last_modified_at | date: '%s' %}
+  {% assign diff_date = three_months_ago | minus: page_date %}
+
+  {% if diff_date >= 0 %}
+  <div class=uk-alert-warning uk-card" style="margin-top: 25px !important;" uk-alert>
+    <b>Note:</b>
+    This page has been published more than 6 months ago, and parts
+    of its contents are very likely to be obsolete by now.
```
hstove, line 28: This language is too strong - "Very likely" is an overreach. Most likely, a majority of the 6+ month old pages are still up to date, and there are more recently updated pages (zero to dapp) that are "outdated".

Maybe an option is to increase this threshold to something like 1 year?

I think there would be more value in manually adding a notice like this to specific pages that we know we need to update. And then, we should be tracking those updates somewhere.

agraebe: We could add them manually as well, this would be another workstream that requires us to go through each page and review it. Is that something we want to sign up for right now? I couldn't judge by myself and I assume it requires multiple people going through pages.

We might want to use both approaches, actually...
- make the callout for _really_ old docs, 1 year+ (and soften the language as per feedback above)
- identify only the most relevant pages for review and add the callout manually to these

markmhx: If we're going to automate this, I would recommend softening the language as well to something like "may be partially out of date" (since if a page is _entirely_ out of date, we're likely to have removed it).

agraebe: I hear you both, the language needs to change

agraebe:
@markmhx @hstove should we also identify the top pages that should have a manual callout. Pages that were changed recently but the content is still outdated and it should be noted?

markmhx:
I'm personally not familiar enough with the docs in general at the moment to know which pages require that kind of manual callout, so it's hard for me to say it's a good practice to apply here. Do you have specific ones in mind?

agraebe:
@hstove this is ready for review. One hour callout with improved language based on feedback. Skipping the manual review for now - part of that was done in the information architecture PR

agraebe:
Seems to be irrelevant right now

PR #540, opened 2020-04-13
Docs updates relating to testnet

hstove:
- Updates the clarity language reference, which is auto-generated from core
- update Jekyll version. This fixed an issue where the docs weren't auto-reloading after I would update a page
- Added some docs for `@blockstack/connect`


_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
+---
+layout: learn
+permalink: /:collection/:path.html
+---
+# Get Started
+{:.no_toc}
+
+Blockstack Connect is a Javascript library for integrating your application with Stacks v2. With Connect, you get some big benefits:
+
+<!-- -  -->
```
agraebe, line 10: @markmhx could we get the top benefits of Connect for devs listed here?

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
+---
+layout: learn
+permalink: /:collection/:path.html
+---
+# Get Started
+{:.no_toc}
+
+Blockstack Connect is a Javascript library for integrating your application with Stacks v2. With Connect, you get some big benefits:
```
agraebe, line 8: (FYI - for the rest of comments on here, I'll take on a "beginners mindset". assuming the reading is lacking sufficient context)

What's Stacks v2? Do I want to integrate it?

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+With npm:
+
+```bash
+npm install --save @blockstack/connect
+```
+
+## Usage
+
```
agraebe, line 30: Maybe an overview of the 4 sub sections would help. Not sure if starting with AuthOptions without more context will be understandable

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+  sendToSignIn?: boolean;
+  userSession?: UserSession;
+  appDetails: {
+    name: string;
+    icon: string;
+  };
+}
+```
+
+- `redirectTo`: The path in your app where users go after sign in.
```
agraebe, line 53: Might be easier to read as a table

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+- `appDetails`: an optional object which includes `appName: string` and `appIcon: string`. This will speed up the process of loading your app's information during onboarding.
+- `manifestPath`: __(optional)__ - the path in your app where your manifest.json file can be found
+- `finished`: __(optional)__ - A callback that can be invoked after authentication. This prevents having to do a whole page refresh in a new tab. One argument is passed to this callback, which is an object with `userSession` included. If included, then the `redirectTo` path is ignored, and the user will be logged in automatically.
+- `authOrigin`: __(optional)__ - The URL you'd like to use for authentication. Only necessary for local development of the authenticator.
```
agraebe, line 57: where/how to get the local url+port? any docs somewhere helping with that?

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+    name: string;
+    icon: string;
+  };
+}
+```
+
+- `redirectTo`: The path in your app where users go after sign in.
+- `appDetails`: an optional object which includes `appName: string` and `appIcon: string`. This will speed up the process of loading your app's information during onboarding.
+- `manifestPath`: __(optional)__ - the path in your app where your manifest.json file can be found
+- `finished`: __(optional)__ - A callback that can be invoked after authentication. This prevents having to do a whole page refresh in a new tab. One argument is passed to this callback, which is an object with `userSession` included. If included, then the `redirectTo` path is ignored, and the user will be logged in automatically.
```
agraebe, line 56: is `redirectTo` technically also optional?

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+- `authOrigin`: __(optional)__ - The URL you'd like to use for authentication. Only necessary for local development of the authenticator.
+- `sendToSignIn`: __(optional)__ - defaults to `false`. Whether the user should go straight to the 'sign in' flow.
```
agraebe, line 58: What is the alternative? In what cases is this relevant?

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+- `userSession`: __(optional)__ - pass a `UserSession` instance to use for authentication. If it's not passed, `@blockstack/connect` will create one for you.
+
+### In React Apps
+
```
agraebe, line 62: I would be confused by `infrastructure`. maybe a lighter term usable?

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+If you're using `connect` in a React app, then the best option is to include `connect`'s React infrastructure and hooks in your React app.
+
+First, setup the `Connect` provider at the "top-level" of your app - probably next to wherever you would put a Redux provider, for example.
```
agraebe, line 65: I'd maybe make this a sub chapter > "setup auth"

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+};
+
+const App = () => (
+  <Connect authOptions={authOptions}>
+    // the rest of your app's components
+  </Connect>
+)
+```
+
+Later, when you want to begin the onboarding process, use the `useConnect` hook to get `connect`'s `doOpenAuth` method.
```
agraebe, line 88: I'd maybe make this a sub chapter > "sign in"

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+### In ES6 (non-React) apps
+
+If you aren't using React, or just want a simpler API, then you can use the `showBlockstackConnect` method.
+
+
+```javascript
+import { showBlockstackConnect } from '@blockstack/connect';
+
+const authOptions = { /** See docs above for options */ };
+showBlockstackConnect(authOptions);
```
agraebe, line 113: equivalent to 2 sub chapters above ... how would sign in look like?

_develop/connect/get-started.md
```diff
@@ -0,0 +1,138 @@
...
+
+  return (
+    <Button onClick={doOpenAuth}>
+      Sign In
+    </Button>
+  )
+}
+```
+
+### In ES6 (non-React) apps
```
agraebe, line 104: I'm a bit confused as to how this relates to sign in methods provided by http://blockstack.github.io/blockstack.js/index.html...

Do we want to sunset/replace auth in blockstack.js? @markmhx

_develop/connect/overview.md
```diff
@@ -0,0 +1,25 @@
...
+# Overview
+{:.no_toc}
+
+* TOC
+{:toc}
+
+Blockstack Connect is a Javascript library for integrating your application with Stacks v2. With Connect, you get some big benefits:
+
+- Pre-built UI to introduce your users to Blockstack, before they authenticate
+- Integration with our new authenticator
```
agraebe, line 14: what's the end benefit of that?

_develop/connect/overview.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Demo
+
+To get a test feel for the user experience of using Connect, you can use [Banter](https://banter.pub).
+
+## How does this compare to `blockstack.js`?
+
+Although [`blockstack.js`](https://github.com/blockstack/blockstack.js) exposes everything you need to handle authentication with Blockstack, there is still the hard problem of getting users used to the paradigm of authentication that is privacy-first and self-sovereign. Many apps implement their own dialogues before authentication, which explain what Blockstack is and why they use it.
+
+`@blockstack/connect` provides developers with a plug-and-play API that is simple to use, and provides great, out-of-the-box education that end-users can understand.
```
agraebe, line 25: I don't feel like the question is answered? Why is this connect mechanism not in blockstack.js for a good developer experience? Two competing auth methods shouldn't be there. The mental overhead to implement auth in Blockstack just went up significantly I think

_develop/connect/use-with-clarity.md
```diff
@@ -0,0 +1,92 @@
...
+{:.no_toc}
+
+With Connect, you can interact with the Stacks 2.0 blockchain. You can allow your users to send transactions and interact with smart contracts.
+
+* TOC
+{:toc}
+
+## How it works
+
+For your app's users to be able to execute a smart contract function, they need to sign and broadcast a transaction. It's important that users remain control of the private keys that sign these transactions. Connect provides an easy-to-use workflow that allows your users to securely sign transactions.
```
agraebe, line 15: maybe an animation/preview would help here? I think that would make it tangible

_develop/connect/use-with-clarity.md
```diff
@@ -0,0 +1,92 @@
...
+
+Connect allows you to open the authenticator with parameters indicating the details of the transaction - like the smart contract address, function name, and specific arguments. Your users get the chance to see these details, and then sign and broadcast the transaction in a single click. Their transaction will be securely signed and broadcasted onto the Stacks blockchain. After this is done, a callback is fired to allow you to update your app.
```
agraebe, line 17: "the authenticator" is likely not a term/reference readers would get. maybe add a super brief explanation or just skip introducing the term?

_develop/connect/use-with-clarity.md
```diff
@@ -0,0 +1,92 @@
...
+
+## API
```
agraebe, line 19: Can we call this something else? I feel like this might reason people to think of web APIs. How about just interface? Or Usage?

_develop/connect/use-with-clarity.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+- `contractAddress`: The Stacks address that published this contract
+- `contractName`: The name that was used when publishing this contract
+- `functionName`: The name of the function you're calling. This needs to be a [public function](/core/smart/clarityRef.html#define-public).
+- `functionArgs`: The arguments you're calling the function with
+- `authOrigin`: _(optional)_ The URL for the authenticator. You only need to use this if you're developing the authenticator locally. Do not use in production
+- `userSession`: _(optional)_ A `UserSession` instance
+- `appDetails`: A dictionary that includes `name` and `icon`
+- `finished`: A callback that is fired when the transaction is signed and broadcasted. Your callback will receive an object back with a `txId` and a `txRaw`.
```
agraebe, line 45: how does txRaw look like?

_develop/connect/use-with-clarity.md
```diff
@@ -0,0 +1,92 @@
...
+  userSession?: UserSession;
+  appDetails?: AuthOptions['appDetails'];
+  finished?: (data: FinishedTxData) => void;
+}
+```
+
+- `contractAddress`: The Stacks address that published this contract
+- `contractName`: The name that was used when publishing this contract
+- `functionName`: The name of the function you're calling. This needs to be a [public function](/core/smart/clarityRef.html#define-public).
+- `functionArgs`: The arguments you're calling the function with
```
agraebe, line 41: isn't this optional based on the interface above?

_develop/connect/overview.md
```diff
@@ -0,0 +1,25 @@
...
+layout: learn
+permalink: /:collection/:path.html
+---
+# Overview
+{:.no_toc}
+
+* TOC
+{:toc}
+
+Blockstack Connect is a Javascript library for integrating your application with Stacks v2. With Connect, you get some big benefits:
```
kyranjamie, line 11: capital s in JavaScript

_develop/connect/overview.md
```diff
@@ -0,0 +1,25 @@
...
+- Integration with our new authenticator
+- Ability to interact with the Stacks 2.0 blockchain, with support for smart contracts and other transactions.
+
+## Demo
+
+To get a test feel for the user experience of using Connect, you can use [Banter](https://banter.pub).
+
+## How does this compare to `blockstack.js`?
+
+Although [`blockstack.js`](https://github.com/blockstack/blockstack.js) exposes everything you need to handle authentication with Blockstack, there is still the hard problem of getting users used to the paradigm of authentication that is privacy-first and self-sovereign. Many apps implement their own dialogues before authentication, which explain what Blockstack is and why they use it.
```
kyranjamie, line 23: maybe `users used to the` --> `users familiar with the`

_develop/connect/use-with-clarity.md
```diff
@@ -0,0 +1,92 @@
...
+- `finished`: A callback that is fired when the transaction is signed and broadcasted. Your callback will receive an object back with a `txId` and a `txRaw`.
+
+### Usage in ES6 (non-React) apps
+
+```ts
+import { openContractCall } from '@blockstack/connect';
+
+// Here's an example of options:
+const options = {
+  contractAddress: 'ST22T6ZS7HVWEMZHHFK77H4GTNDTWNPQAX8WZAKHJ',
```
kyranjamie, line 54: Just a thought: should Blockstack create a handful of test/example/demo stacks addresses for which the org holds the keys. 

This'd be one layer of protection again some silly mistakes.

_develop/connect/overview.md
hstove: Better conversion rates, higher education rates

_develop/connect/get-started.md
hstove: I don't think so - Connect uses those APIs. This is more of a 'wrapper' around Blockstack auth, with added UI and the ability to execute a callback after auth (instead of only allowing redirects).

_develop/connect/get-started.md
hstove: Alternative is the user gets sent through the sign up flow.

_develop/connect/get-started.md
hstove: If you're developing the authenticator locally, it's localhost:8080. Docs for that are in https://github.com/blockstack/ux/blob/master/packages/app/README.md

_develop/connect/use-with-clarity.md
hstove: Yeah, not a terrible idea!

hstove: The pro move would be to auto-siphon any mainnet funds from those addresses into a dev fund :p

kyranjamie: Hank's "burnt" crypto address

_develop/connect/use-with-clarity.md
hstove: Added a note that these are both strings :+1:

_develop/connect/use-with-clarity.md
hstove: Yes, updated in table

_develop/connect/use-with-clarity.md
hstove: Changed to `Usage`, agreed that's better :+1:

_develop/connect/use-with-clarity.md
hstove: Agreed - @markmhx can we turn that demo into a Gif or something?

_develop/connect/overview.md
hstove: Main reason is that Connect includes UI, which bundles React. We don't want to include all of that into blockstack.js. Better might be to include the auth methods in some light library or blockstack.js, and change all docs to _only_ point to that.

_develop/connect/use-with-clarity.md
hstove: Yeah - I struggled with how to phrase this. Agreed it doesn't mean anything. We need to settle on a proper name. @markmhx

_develop/connect/get-started.md
hstove: Changed to `React Provider`, which is a pretty well-known concept for React devs.

_develop/connect/get-started.md
hstove: FYI I removed that param from the docs - since it's mostly internal, we don't need to add the mental overhead here.

_develop/connect/use-with-clarity.md
agraebe: For the time being, maybe just say `Connect allows you to authenticate users with set parameters. The details of the transaction will be shown on the UI`

_develop/connect/overview.md
agraebe: I'd love to chat about that. I don't yet understand if we're now replacing the auth methods in blockstack.js? If so, we have to deprecate them. More options for the same goals are confusing.

On a higher level, I believe devs liked that there's just one simple lib to use. This will now be different, right? Let's take this discussion offline - outside of this scope :)

_develop/connect/overview.md
agraebe: Would write that instead - "new authenticator" isn't necessarily appealing to someone who isn't familiar with the old one

_develop/connect/get-started.md
agraebe: Maybe add that: `Whether the user should go straight to the 'sign in' flow (false) or be presented with the 'sign up' flow (true) instead.`?

_develop/connect/get-started.md
markmhx: @hstove and @aulneau may be best for articulating these from a developer perspective

markmhx: Although it seems we want the same content here as here: https://github.com/blockstack/docs.blockstack/pull/540#pullrequestreview-392416768

_develop/connect/overview.md
markmhx: "Easily onboard new users and authenticate existing ones with plug-and-play Secret Key generation and username management"

_develop/connect/use-with-clarity.md
markmhx: Sounds like a good idea generally, but which demo do you have in mind exactly?

_develop/connect/use-with-clarity.md
markmhx: That works for me. But if we find ourselves reaching for "the authenticator" in our documentation, perhaps we should explain upfront somewhere that "The Blockstack authenticator is a hosted web app that generates cryptographically secure Secret Keys for your new users so they can access your app and manage their data in a secure manner."

_develop/connect/use-with-clarity.md
hstove: I was thinking the demo from the forum post that introduces Secret Key

_develop/connect/get-started.md
hstove: Thanks @agraebe , great suggestion! I've added that.

hstove:
@agraebe @markmhx I've started a first draft for running a Neon miner, given @lgalabru 's guidance this morning.

It should be readable at https://deploy-preview-540--docsproduction.netlify.com/core/smart/neon-node.html (if the deploy works)

markmhx:
We need to update the Clarity docs to refer to the staged authenticator instead of prod until we have transaction signing ready for prod.

markmhx:
Did we want to push the node tutorial live given it has TODO sections still marked for completion? https://docs.blockstack.org/core/smart/neon-node.html

markmhx:
Separately, as mentioned in my review notes shared with @hstove and @agraebe via Paper, I got blocked trying the Clarity "Hello World" tutorial when `npm init clarity-starter` because the template wasn't available.

hstove:
Yes, intentional on adding the neon-node tutorial - it's not linked anywhere in the nav, so not discoverable ATM.

PR #541, opened 2020-04-14
Updating Clarity docs and  tutorials

agraebe:
Includes the following changes:

- [x] Docs overview section update
- [x] "Core" docs section information architecture update
- [x] CSS fixes for better readability 
- [x] Beginner contract tutorial: Hello world
- [x] Intermediate contract tutorial: Counter
- [x] "Clarity" docs section updates


_core/smart/clarityRef.md
```diff
@@ -18,7 +18,7 @@ The `get-block-info` function fetches property details for a block at a specifie
 (get-block-info time 10) ;; Returns 1557860301
 ```
 
-Because the Clarity language is in pre-release, the block properties that are fetched are simulated properties from a SQLite database. The available property names are:
+Because the Clarity language is in early release, the block properties that are fetched are simulated properties from a SQLite database. The available property names are:
```
hstove, line 21: I think we should just remove this whole "Block properties" section. It's not really helpful to a developer, and it's definitely not relevant to the language reference of Clarity.

_data/navigation_home.yml
```diff
@@ -1,24 +1,24 @@
...
-  icon: info
-  doc: org/overview
-
-- title: Use Blockchain Applications
-  desc: Learn about the New Internet and blockchain applications. Create an identity and learn how to use it.
-  icon: settings
-  doc: browser/browser-introduction
-
-- title: Build Blockchain Applications
+- title: Build Decentralized Apps
```
hstove, line 3: I like these homepage section updates!

_core/smart/clarityRef.md
agraebe: done!

_core/smart/tutorial.md
```diff
@@ -3,306 +3,207 @@ layout: core
...
+    Clarity is a programming language based on [LISP](https://en.wikipedia.org/wiki/Lisp_(programming_language)). Most notably, Clarity is designed for static analysis, not compiled, and **not** [Turing complete](https://en.wikipedia.org/wiki/Turing_completeness).
 
-3. Type check the `tokens.clar` contract, it should pass a check as it does not use the `contract-call` function:
+    Let's go through the source code. Notice how the program and each statement is enclosed in `()` (parentheses). You'll see that the smart contract consists of two public methods. Starting at the top, let's review line by line:
 
-    ```bash
-    # clarity-cli check sample-programs/tokens.clar /data/db
-    Checks passed.
+    ```cl
+    (define-public (say-hi)
```
hstove, line 91: I know we don't have any style guides for Clarity, but I'm not a fan of this indentation. I think this is much more readable:

```racket
(define-public (say-hi)
  (ok "hello world")
)
```

Mainly for functions longer than one line

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,193 @@
...
+
+    ```bash
+    counter contract test suite
+        ✓ should have a valid syntax
+        deploying an instance of the contract
+        1) should start at zero
+        2) should increment
+        3) should decrement
+
+    hello world contract test suite
```
hstove, line 65: Need to remove this section. I think this is coming from a local existing package, because the "hello world" tests should not be there. The test results should be "1 passing, 3 failing"

_core/smart/overview.md
```diff
@@ -1,135 +1,45 @@
...
 
 * The only atomic types are booleans, integers, fixed length buffers, and principals
 * Recursion is illegal and there is no lambda function.
 * Looping may only be performed via `map`, `filter`, or `fold`
 * There is support for lists of the atomic types, however, the only variable length lists in the language appear as function inputs; There is no support for list operations like append or join.
 * Variables are created via `let` binding and there is no support for mutating functions like `set`.
 
+## Learning Clarity
+
+You can try a [Hello World tutorial](tutorial-counter.html) or jump right into the [language reference](clarityRef.html).
```
hstove, line 45: Wrong link for "hello world tutorial"

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,193 @@
...
+
+    You should be familiar with the test set up from the Hello World tutorial. Notice how the instance of the smart contract is created on line 8:
+
+    ```js
+    counterClient = new Client("SP3GWX3NE58KXHESRYE4DYQ1S31PQJTCRXB3PE9SB.counter", "counter", provider);
+    ```
+
+    That tells us that the new smart contract is named `counter` and that we need to start by creating a new file for the smart contract: `contracts/counter.clar`. Note that the `contracts` folder is assumed as the base folder and that every Clarity file has the suffix `.clar`.
+
+2. Let's create the new file:
```
hstove, line 96: The file is already created, it's just blank

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,193 @@
...
+
+    ```shell
+    touch contracts/counter.clar
+    ```
+
+3. With the editor of your choice, open the file and add the following lines of code:
+
+    ```cl
+    (define-data-var counter int 0)
```
hstove, line 105: The indentation here is throwing things off, and copy/paste results in wacky indentation. There is no need to indent the whole code block, the 3 backticks is all that is required.

_core/smart/overview.md
agraebe: Fixed

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,193 @@
...
+    ```
+
+    You should now only see 2 failing tests! `should start at zero` is passing, and you successfully build your first part of the contract. Congrats!
+
+    However, we don't stop here. Let's implement increment and decrement methods.
+
+5. Add the following lines to the bottom of the `counter.clar` file and take a few seconds to review them:
+
+    ```cl
+    (define-public (increment)
```
hstove, line 131: Again, we don't have formal style guides yet, but most/all of the Clarity code I've seen so far uses 2-space indentation. It's also what we use in the hello-world starter. Would recommend switching to 2 spaces

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,193 @@
...
+
+    ```shell
+    counter contract test suite
+        ✓ should have a valid syntax
+        deploying an instance of the contract
+        ✓ should start at zero
+        ✓ should increment (95ms)
+        ✓ should decrement (92ms)
+
+    hello world contract test suite
```
hstove, line 155: Also need to remove this section of test output

_core/smart/tutorial-counter.md
agraebe: Tried that, but it creates a but with the numbering (3. will become 1. at every paragraph)

hstove: D'oh... yeah that makes sense. Can we try wrapping the line in `<p>..</p>`?

hstove:
I've added some small nits, but I've gone through these to test them out, and it works very well. This is great - the contents are in good sizes, not too long or short. And it's pretty simple to get started, especially if you've already installed `npm`.

_core/smart/tutorial-counter.md
agraebe: Great!

_core/smart/tutorial-counter.md
agraebe: fixed

reedrosenbluth:
Awesome work! Haven't finished looking through this but something that stood out to me was that the formatting of the parentheses in the Clarity code snippets seems off.

In Lisp something like:

```Lisp
(define-public (say-hi)
  (ok "hello world")
)
```

is typically formatted like:

```Lisp
(define-public (say-hi)
  (ok "hello world"))
```

_core/smart/overview.md
```diff
@@ -1,135 +1,55 @@
...
-As the name implies, the `check` ensures the contract definition passes a type check; passing programs will returns an exit code of `0` (zero). Once a contract passes a check, you `launch` it.
-
-```bash
-root@4224dd95b5f5:/data# clarity-cli launch hello ./hello.clar /data/db
-Buffer(BuffData { data: [104, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100] })
-Contract initialized!
-```
-
-Because Clarity does not support simple strings, it stores the `hello world` string in a buffer. Printing out that string displays the ASCII representation for each character. You can see the record of this contract's launch in the corresponding database:
+* The language is not intended to be compiled
```
jcnelson, line 36: I think affirmative wording would sound better here, since we don't want to be framing our key features to sound like limitations.  We've used `The language is interpreted` and `The language is decidable` as alternatives in most other places (i.e. in talks, blog posts, tweets, etc.).

_core/smart/overview.md
```diff
@@ -1,135 +1,55 @@
...
+* The language is not Turing complete
+  
+These differences allow for static analysis of programs to determine properties like runtime cost and data usage. Omitting compilation prevents the possibility of error or bugs introduced at the compiler level.
```
jcnelson, line 39: "In addition, using an interpreted language ensures that the code that executes is human-readable, and is auditable by users."

_core/smart/overview.md
```diff
@@ -1,135 +1,55 @@
...
 
-```sql
-sqlite> select * from contracts;
-1|hello|{"contract_context":{"name":"hello","variables":{},"functions":{}}}
-sqlite> select * from type_analysis_table;
-1|hello|{"private_function_types":{},"variable_types":{},"public_function_types":{},"read_only_function_types":{},"map_types":{}}
-sqlite> 
-```
+A Clarity smart contract is composed of two parts &mdash; a data-space and a set of functions. Only the associated smart contract may modify its corresponding data-space on the Blockchain. Functions are private unless they are defined as public functions. Users call smart contracts' public functions by broadcasting a transaction on the Blockchain which invokes the public function.
```
jcnelson, line 41: "Functions are private unless they are defined as public functions" is a tautology.

Recommend something like "Functions may be private and thus callable only from within the smart contract, or public and thus callable from other contracts."

_core/smart/overview.md
```diff
@@ -1,135 +1,55 @@
...
 
-## Language rules and limitations
+Contracts can also call public functions from other smart contracts. The ability to do a static analysis of a smart contract allows a user to determine dependency between contracts.
```
jcnelson, line 43: This isn't due to anything special in Clarity -- you can determine a transaction's call graph in Solidity as well (albeit doing so requires _running_ the code).

I think this is meant to read something like "The fact that Clarity is a decidable language means that users can determine the set of _reachable_ code from any function they want to call."

_core/smart/overview.md
```diff
@@ -1,135 +1,55 @@
...
 
-The Clarity smart contract has the following limitations:
+Note some of the key Clarity language rules and limitations:
 
 * The only atomic types are booleans, integers, fixed length buffers, and principals
 * Recursion is illegal and there is no lambda function.
 * Looping may only be performed via `map`, `filter`, or `fold`
 * There is support for lists of the atomic types, however, the only variable length lists in the language appear as function inputs; There is no support for list operations like append or join.
```
jcnelson, line 50: There shouldn't be a semicolon here.  It should be a period.

_core/smart/tutorial.md
```diff
@@ -1,308 +1,208 @@
...
+    cat contracts/hello-world.clar
     ```
 
-    When the `check` command executes successfully and exits with the stand UNIX `0` exit code.
-
-4. Generate a demo Stacks address for testing your contract.
+    You should see the contract source code. Take a few seconds to review the content.
 
-   This address is used to name your contract at launch time. You can use any existing Stacks address. For this sample, you are going to use the `generate_address` command to create one.
+    Clarity is a programming language based on [LISP](https://en.wikipedia.org/wiki/Lisp_(programming_language)). Most notably, Clarity is designed for static analysis, not compiled, and **not** [Turing complete](https://en.wikipedia.org/wiki/Turing_completeness).
```
jcnelson, line 81: Same idea here -- we've historically been phrasing Clarity features in the affirmative, since we don't want to frame these key features as if they were limitations.

_core/smart/tutorial.md
```diff
@@ -1,308 +1,208 @@
...
-
-    ```bash
-    # DEMO_ADDRESS=SP28Z69HE5H70BVRG4VGKN4SYNVJ1J0417WVCKZWM
+    (define-public (echo-number (val int))
+      (ok val)
+    )
     ```
 
-6. Launch the `tokens.clar` contract and assign it to your `DEMO_ADDRESS` address.
+    On the first line, a new public function `say-hi` is declared. To create private functions, you would use the `define-private` function. Private functions can only be executed by the current smart contract and not from the outside. Only public functions can be called from outside e.g., through other smart contracts or applications. The need to call another smart contract can have different reasons. It allows to reuse existing smart contracts published by other developers. It may also be required because complex implementations are often split out into a set of contracts to achieve a [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns).
```
jcnelson, line 95: `define-private` is not a function; it's a keyword.

jcnelson: `from outside e.g., through other smart contracts or applications`

might be better phrased as 

`from outside, by other smart contracts.`  Programs running outside the blockchain cannot, in general, invoke code within the blockchain (the limited exception is `define-read-only`).

jcnelson: `The need to call another smart contract can have different reasons. It allows to reuse existing smart contracts published by other developers. It may also be required because complex implementations are often split out into a set of contracts to achieve a [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns).`

might be better phrased as 

`The reason public functions exist is to enable re-using code that is already available in other smart contracts, and to enable developers to break complex smart contracts into smaller, simpler smart contracts (an exercise in [separating concerns](https://en.wikipedia.org/wiki/Separation_of_concerns)).`

_core/smart/tutorial.md
```diff
@@ -1,308 +1,208 @@
...
-    # echo "(get-balance '$DEMO_ADDRESS)" | clarity-cli eval $DEMO_ADDRESS.tokens /data/db
-    Program executed successfully! Output: 
-    100000
-    ```
+provider = await ProviderRegistry.createProvider();
+helloWorldClient = new Client("SP3GWX3NE58KXHESRYE4DYQ1S31PQJTCRXB3PE9SB.hello-world", "hello-world", provider);
+```
 
-## Task 6: Spend tokens by registering a name 
+Take a look at the client initialization. It requires a contract id and name in the following format: `{owner_stacks_address}.{contract_identifier}`. The second parameter indicates the location of the smart contract file, without the `.clar` suffix. By default, the location is assumed to be relative to the `contracts` folder.
```
jcnelson, line 165: `The second parameter` should be `The second field`.

_core/smart/tutorial.md
```diff
@@ -1,308 +1,208 @@
...
 
-Now, let's register a name using the `names.clar` contract. Names can _only_ be integers in this sample contract, so you'll register the name 10 in this environment.
+As you can see above, a sample stacks address and contract identifier is already provided for you. You don't need to modify anything.
```
jcnelson, line 167: `stacks` should be capitalized.

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,177 @@
...
+
+    ```cl
+    (define-data-var counter int 0)
+
+    (define-public (get-counter)
+      (ok (var-get counter))
+    )
+    ```
+
+    The first line initializes a new integer variable `counter` with the value set to `0` using the [`define-data-var`](https://docs.blockstack.org/core/smart/clarityref#define-data-var) statement. It is important to note that all definition statements in Clarity need to be at the top of the file.
```
jcnelson, line 102: `It is important to note that all definition statements in Clarity need to be at the top of the file.`

I'm not sure this is true anymore.  @kantai or @lgalabru would know for sure.

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,177 @@
...
+
+    To provide access to the `counter` variable from outside of the current smart contract, we need to declare a public get function. The last lines of the code add a public `get-counter` function. The [`var-get`](https://docs.blockstack.org/core/smart/clarityref#var-get) statement looks for a variable in the contract's data map and returns it.
```
jcnelson, line 104: `declare a public get function` --> `declare a public function to get it`.

jcnelson: Also, `data map` --> `data space`.  We need to be careful with the word `map`, since Clarity has a notion of data variables and data maps (both are part of the data space of a contract).

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,177 @@
...
+    1) should start at zero
+    2) should increment
+    3) should decrement
+
+
+1 passing (734ms)
+3 failing
+```
+
+It looks like we see some failed test! That is on purpose - we will implement the new smart contract in the next steps! After every increment of the contract, we will rerun the tests to ensure we're on the right track.
```
jcnelson, line 70: `test!` should be `tests!`

jcnelson: Recommend `After every increment of the contract` --> `After every step in this tutorial`

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+You have to select a template and a name for your local folder. For the counter template used in this tutorial, ensure to type `counter` and hit ENTER:
+
+```bash
+? Template - one of [hello-world, counter]: counter
+? Project name: (clarity-counter)
+```
+
+Finally, the project dependencies are installed and your project is ready for development. Because you already completed the Hello World tutorial, the project structure is familiar to you. The main difference is that we have additional tests for a new counter smart contract.
```
jcnelson, line 43: Should we quote `Hello World`?

_core/smart/tutorial-counter.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Step 2: Running tests
+
+Smart contracts are often developed in a test-driven approach to ensure code quality but also to speed up the development cycle by removing the need to push every change to the Blockchain before executing it. We will do the same in this project. Now, let's run the tests and review the results:
```
jcnelson, line 47: `Blockchain` should not be capitalized.

Also, recommend wording this sentence as `Smart contracts are often developed in a test-driven approach.  This not only improves code quality, but also removes the need to push every iteration to the blockchain before executing it.`

_core/smart/tutorial-counter.md
kantai: It matters during evaluation -- a contract is executed in order. While functions may be defined after another function that calls them, they cannot follow statements that invoke them during the contract initialization, e.g.:

```
(define-private (foo)
   (bar))
(define-private (bar)
   1)
(foo)
```

Is fine.

```
(foo)
(define-private (foo)
   (bar))
(define-private (bar)
   1)
```

Will yield a runtime error.

GinaAbrams:
I like that the use cases includes App Chains, but if I were an outsider I might not know what that means from just those two words... or do we expect all users to grok that? 🤔  

Would `App Chains: app specific blockchains` potentially help in this instance, or sound equally as vague @jcnelson ?

_core/smart/overview.md
```diff
@@ -1,135 +1,55 @@
...
 
-<div class="uk-card uk-card-default uk-card-body">
-<h5>Clarity is in pre-release</h5>
-<p>Clarity, its accompanying toolset, and the SDK are in pre-release.  If you encounter issues with or have feature requests regarding Clarity, please create an issue on the <a href='https://github.com/blockstack/blockstack-core/issues' target='_blank'>blockstack/blockstack-core</a> repository. To read previous or join ongoing discussions about smart contracts in general and Clarity in particular, visit the <strong><a href='https://forum.blockstack.org/c/clarity' target='_blank'>Smart Contracts</a></strong> topic in the Blockstack Forum.
-</p>
-</div>
+## Smart contracts
 
-## Who should use smart contracts?
+Smart contracts allow two parties to exchange anything of value (money, property, shares), in an automated, auditable, and secure way _without the services of a middleman_. Nick Szabo introduced the canonical metaphor for smart contracts, a vending machine.
```
kantai, line 16: I'd be in favor of a more generic definition of a smart contract here. They enable much more than just exchanging things of value, and could be introduced in that way. Something like:


Smart contracts are programs that can be executed in a transparent and verifiable fashion. When deployed to a blockchain, they allow parties to submit transactions to the blockchain which execute those programs. Such smart contract operation is automated, auditable, and consistent.

Smart contracts enable parties to exchange things of value in an automated and secure way. For example, a "swap contract" could allow two parties to exchange two different assets without trusting one another. One party would initiate the swap by submitting a transaction to transfer an asset to the contract. Then, when the second party transferred the second asset, the smart contract would distribute the assets to each party. If the second party never transferred the second asset, the smart contract would release the first asset back to the first part.

_core/smart/overview.md
kantai: Yeah -- I'd probably phrase this as

* ~The language is not intended to be compiled.~ The language is designed to be broadcast on the blockchain as is.
* ~The language is not Turing complete~ The language is decidable.

_core/smart/overview.md
kantai: Yep -- I think this was language from when `(define (foo))` defined a private function, whereas now, `(define-private ...)` is more explicit.

_core/smart/overview.md
agraebe: Love it

_core/smart/overview.md
agraebe: Cool, will update

_core/smart/tutorial-counter.md
agraebe: yes, thanks!

_core/smart/tutorial-counter.md
agraebe: love it!

_core/smart/overview.md
jcnelson: I was going back and forth about this example as well.

Do you think there's a way we can frame smart contracts as a tool for enabling a user-owned internet?  I think we're doing ourselves a disservice trying to talk about smart contracts in terms of programmable money (and nothing else).  An example I was toying with last night looks like:

`Smart contracts are programs that run on a blockchain.  In Blockstack, the main use-case for smart contracts is to manage _global shared state_ for an app.  This is data that is collectively owned and curated by all of an app's users.  For example, the list of all subreddits is global shared state in Reddit.  As another example, the catalog of movies is global shared state in YouTube.  Every user can see and use this data when the use the app, and every user can propose changes to this data, but the data itself belongs to everyone.`

`Smart contracts encode and enforce the rules for curating this data, and are often used to ensure that users can exchange valuable data in an automated and secure way.  For example, a smart contract could allow anyone to add a new movie to a movie catalog, but require that someone first pay for a decryption key before anyone can view it.  To do this, the smart contract could be written such that the movie creator had to publicly disclose the decryption key in order to receive payment.  Such a system would allow movie creators to make money by adding movies to the movie catalog that people want to watch.`

`Because smart contracts run on top of a blockchain, anyone can query them, and anyone can submit transactions to run their code.  The blockchain stores the history of all accepted transactions, so anyone can audit the blockchain in order to independently verify that an app's global shared state has been managed correctly according to the smart contract's rules.`

kantai: Even better!

PR #544, opened 2020-04-20
CLI wallet quickstart

yknl:
This adds a quickstart guide for the CLI wallet with instructions specific to Testnet. The guide will show developers how to install the CLI, create a wallet, receive tokens from the faucet and send tokens out from their account.

The guide can be found under Build Smart Contracts -> References

### DIFF
--- a/_core/smart/cli-wallet-quickstart.md
+++ b/_core/smart/cli-wallet-quickstart.md
@@ -69,9 +69,9 @@ In order to send tokens, we will need 5 parameters.
 
 **Recipient Address** - The Stacks address of the recipient
 
-**Amount** - The number of Stacks to send denoted in microstacks (1 STX = 1000000 mSTX)
+**Amount** - The number of Stacks to send denoted in microstacks (1 STX = 1000000 microstacks)
 
-**Fee Rate** - The transaction fee you are willing to pay send this transaction. Denoted in microstacks per byte. For Testnet use, you can simply set a fee rate of 1.
+**Fee Rate** - The transaction fee rate for this transaction. You can safely set a fee rate of 1 for Testnet.
 
 **Nonce** - The nonce is a number that needs to be incremented monotonically for each transaction from the account. This ensures transactions are not duplicated.
 
--- a/_core/smart/cli-wallet-quickstart.md
+++ b/_core/smart/cli-wallet-quickstart.md
@@ -1,0 +1,99 @@
+---
+layout: smart
+description: "CLI Wallet Guide"
+permalink: /:collection/:path.html
+---
+# Blockstack CLI Wallet Quickstart (Testnet)
+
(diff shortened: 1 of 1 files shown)


_core/smart/cli-wallet-quickstart.md
```diff
@@ -0,0 +1,99 @@
...
+The Stacks address associated with this wallet is: 
+`ST1BG7MHW2R524WMF7X8PGG3V45ZN040EB9EW0GQJ`
+
+Note that this is a Testnet address for use with Testnet only. You will use this address to receive tokens from the Testnet faucet or other people.
+
+The private key is what you will need to send tokens later. 
+
+## Getting Tokens
+
+To receive Testnet Stacks tokens, visit the Testnet stacks token [faucet](https://www.blockstack.org/faucet). Enter the Stacks address you generated above into the address field on the faucet page.
```
agraebe, line 48: is this the confirmed final URL?

_core/smart/cli-wallet-quickstart.md
```diff
@@ -0,0 +1,99 @@
...
+
+Verify that your account has tokens before continuing to the next step to send tokens.
+
+## Sending Tokens
+
+In order to send tokens, we will need 5 parameters. 
+
+**Recipient Address** - The Stacks address of the recipient
+
+**Amount** - The number of Stacks to send denoted in microstacks (1 STX = 1000000 mSTX)
```
agraebe, line 72: If I recall correctly, I've seen two ways to shorten micro STX = uSTX, and mSTX. Maybe just avoid shorten it and write micro Stacks?

jeffdomke: I've found `uSTX` confusing. Sounds like we need to settle on a standard and would like to vote for `mSTX`

aulneau: I vote for uSTX -> https://en.wikipedia.org/wiki/Micro-

_core/smart/cli-wallet-quickstart.md
```diff
@@ -0,0 +1,99 @@
...
+
+**Fee Rate** - The transaction fee you are willing to pay send this transaction. Denoted in microstacks per byte. For Testnet use, you can simply set a fee rate of 1.
```
jeffdomke, line 74: Found this a little cryptic. Not sure if this is the best place for teaching this concept but I might start asking "So the bytes we are measuring are...? All the code below? Something else at runtime?..." Think it is fine to skip but maybe could be a little more descriptive.

agraebe: +1

_core/smart/cli-wallet-quickstart.md
yknl: Since we don't have consensus on this, I'll leave it as microstacks for now.

_core/smart/cli-wallet-quickstart.md
yknl: No but we can quickly update this once the faucet URL is finalized.

PR #550, opened 2020-04-21
Incorporate feedback from Mark

agraebe:
Will wait for input from @jcnelson on:
- Principals language
- Movie catalog example

### DIFF
--- a/_core/smart/overview.md
+++ b/_core/smart/overview.md
@@ -13,9 +13,13 @@ Clarity is a smart contracting language for use with the Stacks 2.0 blockchain. 
 
 ## Smart contracts
 
-Smart contracts encode and enforce the rules for data curation, and are often used to ensure that users can exchange valuable data in an programmatic and secure way. For example, a smart contract could allow creators to add a new movie to a movie catalog, but require that users first pay for a decryption key before viewing it. To do this, the smart contract could be written such that the movie creator had to publicly disclose the decryption key in order to receive payment. Such a system would allow movie creators to make money by adding movies to the movie catalog that people want to watch.
+Smart contracts encode and enforce the rules for data curation, and are often used to ensure that users can exchange valuable data in an programmatic and secure way. For example, a smart contract could allow creators to add a new tv show to a streaming service, but require that users first pay for a decryption key before viewing it. To do this, the smart contract could be written such that the movie creator had to publicly disclose the decryption key in order to receive payment. The movie creators would therefore be incentivized to work on tv show episodes that people want to watch, based on the viewers' upfront payments that are locked-in until the creator releases the episode and the decryption key.
 
-Because smart contracts run on top of a blockchain, anyone can query them, and anyone can submit transactions to run their code. Apps use the blockchain to manage a global state that is visible to the public. To get back to ge movie catalog exaple, this state could include a list of user that paid for a specific movie. This is possible because the blockchain stores the history of all accepted transactions. Anyone can audit the blockchain in order to independently verify that an app's global shared state has been managed correctly according to the smart contract's rules.
+Because smart contracts run on top of a blockchain, anyone can query them, and anyone can submit transactions to execute them. A smart contract execution can result in new transactions being written to the blockchain.
+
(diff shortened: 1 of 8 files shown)


_core/smart/overview.md
```diff
@@ -13,31 +13,34 @@ Clarity is a smart contracting language for use with the Stacks 2.0 blockchain.
 
 ## Smart contracts
 
+Smart contracts encode and enforce the rules for data curation, and are often used to ensure that users can exchange valuable data in an programmatic and secure way. For example, a smart contract could allow creators to add a new tv show to a streaming service, but require that users first pay for a decryption key before viewing it. To do this, the smart contract could be written such that the movie creator had to publicly disclose the decryption key in order to receive payment. The movie creators would therefore be incentivized to work on tv show episodes that people want to watch, based on the viewers' upfront payments that are locked-in until the creator releases the episode and the decryption key.
```
hstove, line 16: IMO, not the best example. A smart contract could not determine if they actually made the show or not. Who knows what the "decryption key" is used for? The smart contract could only know if the key is a key for anything - nothing further.

You could tweak this example if we want to stick to it. This could be a DAO, and the customers would have to vote that the movie producer did, in fact, deliver. However, there are lots of further issues than this.

Another option is that the customers are simply pre-purchasing, and they can then use signatures later on when streaming to prove they're a paid customer. However, that's also not a very interesting solution, because it doesn't provide any more value than a centralized service.

_core/smart/overview.md
```diff
@@ -13,31 +13,34 @@ Clarity is a smart contracting language for use with the Stacks 2.0 blockchain.
...
+Because smart contracts run on top of a blockchain, anyone can query them, and anyone can submit transactions to execute them. A smart contract execution can result in new transactions being written to the blockchain.
 
-Because smart contracts run on top of a blockchain, anyone can query them, and anyone can submit transactions to run their code. The blockchain stores the history of all accepted transactions, so anyone can audit the blockchain in order to independently verify that an app's global shared state has been managed correctly according to the smart contract's rules.
+Apps use the blockchain to manage a global state that is visible to the public. To get back to the streaming service exaple, the global state could include a list of user that paid for a specific tv show episoed. This is possible because the blockchain stores the history of all accepted transactions.
```
hstove, line 20: Typo: "exaple" and "episoed"

_core/smart/overview.md
```diff
@@ -13,31 +13,34 @@ Clarity is a smart contracting language for use with the Stacks 2.0 blockchain.
...
+
+Anyone can audit the blockchain in order to independently verify that an app's global shared state has been managed correctly according to the smart contract's rules.
 
 ## Use cases
 
 Not every decentralized application requires smart contracts, but Clarity unlocks interesting capabilities for decentralized applications. Examples of use cases include, but are not limited to:
 
 * Access control (e.g. pay to access)
-* Non-fungible and fungible tokens
+* Non-fungible (e.g. collectibles) and fungible tokens (e.g. stable-coins)
```
hstove, line 29: "Stablecoins" (no hyphen) seems to be the predominant spelling (i.e. wikipedia)

_core/smart/overview.md
hstove: Another problem is that it's not possible for the smart contract to then only give the decryption key to the payers - once it's on the blockchain, anyone can see it.

There are just a lot of complications around the idea of "use smart contracts to pay for digital goods"

agraebe: yeah, that makes sense. do you have a simpler example in mind? something digestible but also exciting =

markmhx: Is "data curation" a commonly used phrase here? Perhaps `Smart contracts encode and enforce the rules for modifying a particular set of data as shared among people and entities who don't necessarily trust each other.` is a better way to state this?

agraebe: Let's continue to iterate on the example but release this one for now

markmhx:
@agraebe I believe this issue can be closed since the feedback was pushed live?

agraebe:
it's not live. this PR needs to be approved by @hstove or someone who owns the repo. @hstove could you approve so I can merge? thanks!

PR #569, opened 2020-04-23
docs: streamlined neon node

hstove:
The blockchain now has default configuration for Neon nodes. There is also a streamlined binary, where you can more easily run a node of any type.

### DIFF
--- a/_core/smart/neon-node.md
+++ b/_core/smart/neon-node.md
@@ -11,7 +11,7 @@ permalink: /:collection/:path.html
 * TOC
 {:toc}
 
-### Download the `stacks-blockchain` repository
+### Download and install the `stacks-blockchain` repository
 
 The first step is to ensure that you have Rust and the support software installed.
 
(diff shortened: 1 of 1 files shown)


_core/smart/neon-node.md
```diff
@@ -31,22 +31,10 @@ git clone https://github.com/blockstack/stacks-blockchain.git
...
-```toml
-bootstrap_node = "048dd4f26101715853533dee005f0915375854fd5be73405f679c1917a5d4d16aaaf3c4c0d7a9c132a36b8c5fe1287f07dad8c910174d789eb24bdfb5ae26f5f27@35.245.47.179:20444"
-```
-
-Next, find the uncommented `peer_host` line in the `[burnchain]` section, and update it to:
-
-```toml
-peer_host = "35.245.47.179"
+```bash
+cargo install --path ./testnet
```
agraebe, line 37: nit: running this takes minutes for me, it's compiling here as well (166 modules).

_core/smart/neon-node.md
```diff
@@ -56,7 +44,7 @@ You're all set to run a node that connects to the Neon network.
 Back in the command line, run:
 
 ```bash
-cargo testnet ./testnet/follower-conf.toml
+stacks-node neon
 ```
 
 The first time you run this, you'll see some logs indicating that the Rust code is being compiled. Once that's done, you should see some logs that look something like the this:
```
agraebe, line 50: I actually now only see one INFO:
```
INFO [1587676136.184] [src/chainstate/stacks/index/marf.rs:732] First-ever block 0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206
```

agraebe: Because the POST and endpoint isn't displayed now, how would people know what endpoint to call in the RPC API? The port is unknown to them, right?

agraebe: Also calling out that there will likely be errors visible in here and it could take a while until a new log shows. might be known to the reader, but just a nit on my side.

hstove:
Merging this because the current prod docs are a bit outdated, and we want to share this in the townhall

PR #571, opened 2020-04-24
Update neon-node.md

pkrasam:
made changes to some of the build and run commands


hstove:
Thanks for the PR! We've already updated this doc with these new commands, so I'm going to close this out.

PR #573, opened 2020-04-24
fix: ensure example command runs

kyranjamie:
Running the command from docs just prints the help text. If we need this .toml file, should we use `start` and pass the config arg? This boots it up for me.

### DIFF
--- a/_core/smart/neon-node.md
+++ b/_core/smart/neon-node.md
@@ -78,9 +78,8 @@ cd ..
 The above code will compile an optimized binary. To use it, run:
 
 ```bash
-./target/release/stacks-node ./testnet/conf/neon-follower-conf.toml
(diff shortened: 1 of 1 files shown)


_core/smart/neon-node.md
```diff
@@ -71,5 +71,5 @@ cd ..
 The above code will compile an optimized binary. To use it, run:
 
 ```bash
-./target/release/stacks-node ./testnet/conf/neon-follower-conf.toml
+./target/release/stacks-node start --config=./testnet/conf/neon-follower-conf.toml
```
agraebe, line 74: @hstove can you please review?

agraebe:
Will merge in #577 , thanks for the PR!

PR #574, opened 2020-04-24
add redirects for zero to dapp URLs

agraebe:
The old URLs will redirect to the Hello, Blockstack tutorial (https://docs.blockstack.org/browser/hello-blockstack.html)


cuevasm:
Thank you!

PR #578, opened 2020-04-28
Update clarity references to latest from stacks-blockchain:master

kantai:
This updates the clarity references to reflect the latest version in stacks-blockchain.


_core/smart/clarityRef.md
```diff
@@ -10,104 +10,126 @@ This file contains the reference for the Clarity language.
...
-To access a named value of a given tuple, the `(get name tuple)` function returns that item from the tuple.
-
-### Optional type
-
-Represents an optional value. This is used in place of the typical usage of "null" values in other languages, and represents a type that can either be some value or `none`. Optional types are used as the return types of data-map functions.
+* Dynamic dispatch: the callee is passed as an argument, and typed
+as a trait reference (<A>).
+
+```scheme
+(define-public (swap (token-a <can-transfer-tokens>)
```
hstove, line 84: One thing I've been curious about - if I was to actually call this function, would the first argument be a fully qualified contract principal?

hstove: Can I also do dynamic dispatch without a trait reference? I.e. `token-a` is just a `principal`. I can still do a `contract-call?` in the function, yeah?

_core/smart/clarityRef.md
```diff
@@ -10,104 +10,126 @@ This file contains the reference for the Clarity language.
...
+    (transfer-from? (principal principal uint) (response uint)))
+```
 
-Response types contain two subtypes -- a response type in the event of `ok` (that is, a public function returns an integer code on success) and an `err` type (that is, a function returns a buffer on error).
+Or imported from an existing contract:
 
-### Algebraic
+```scheme
+(use-trait can-transfer-tokens
+    .contract-defining-trait.can-transfer-tokens)
```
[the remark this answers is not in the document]
kantai: > is this just implying "contract named contract-defining-trait, with the same STX address as this contract"?

Yes -- the `.contract-name` syntax always expands to `this-publisher-address.contract-name`

_core/smart/clarityRef.md
kantai: > One thing I've been curious about - if I was to actually call this function, would the first argument be a fully qualified contract principal?

Yes -- it is either a _literal_ contract principal or a matching trait variable.

> Can I also do dynamic dispatch without a trait reference? I.e. token-a is just a principal. I can still do a contract-call? in the function, yeah?

No -- that's what differentiates the trait execution from general dynamic dispatch. `contract-call` can only be invoked either on a contract-literal or a trait parameter. This ensures that the type system can enforce the invocation during analysis, rather than run time.

PR #581, opened 2020-05-01
fix installing script

jefreybulla:


agraebe:
Thanks for the submission. The install URL is going to stay as is until the Mainnet release (the old CLI without `#feature/stacks-2.0-tx`) is supporting only Stacks 1.0 methods.

Issue #557, opened 2020-04-23
Update "Storage" tutorial with Blockstack Connect


markmhx:
https://docs.blockstack.org/browser/blockstack_storage.html

Update according to the changes made to the underlying app generator.

Issue #556, opened 2020-04-23
Update "Todo List" tutorial with Blockstack Connect


markmhx:
https://docs.blockstack.org/browser/todo-list.html

This tutorial needs its underlying sample application https://github.com/blockstack/blockstack-todos updated to Blockstack Connect and then its written content updated accordingly.

PR #585, opened 2020-05-06
Connect docs for general release

hstove:
This handles #557 and #556. It builds tutorials based off the new app generator, which is available at https://github.com/blockstack/blockstack-app-generator/pull/72


_browser/blockstack_storage.md
```diff
@@ -220,48 +220,36 @@ These are the `UserSession.putFile`, `UserSession.getFile`, and `lookupProfile`
 
 1. Open the `src/components/Profile.js` file.
 
-2. Replace the initial state in the `constructor()` method so that it holds the key properties required by the app.
+2. Replace the initial state in the component method so that it holds the key properties required by the app.
```
agraebe, line 223: I got lost here. What should be replaced? the part in the constructor or the whole component?

_browser/blockstack_storage.md
```diff
@@ -220,48 +220,36 @@ These are the `UserSession.putFile`, `UserSession.getFile`, and `lookupProfile`
...
-            return avatarFallbackImage;
-          },
-        },
-        username: "",
-        newStatus: "",
-        statuses: [],
-        statusIndex: 0,
-        isLoading: false
-      };
+    export const Profile = ({ userData, handleSignOut }) => {
```
agraebe, line 228: This is confusing because we're basically asking to replace the file I guess? It threw me off, especially because the component declaration is different in the file (`export default class Profile extends Component`) - no default, no extension

_browser/blockstack_storage.md
```diff
@@ -220,48 +220,36 @@ These are the `UserSession.putFile`, `UserSession.getFile`, and `lookupProfile`
...
+      const { authOptions } = useConnect();
+      const { userSession } = authOptions;
+      // ...
     }
     ```
 
 
-3. Locate the `render()` method.
-4. Modify the `render()` method to add a text input and submit button to the 
+3. Modify the rendered result to add a text input and submit button to the
```
agraebe, line 242: "to the by replacing it". To the "what" is missing

_browser/blockstack_storage.md
agraebe: I'm also left with a `componentWillMount` method that is now marked as unreachable (on the bottom of the page)

_browser/blockstack_storage.md
```diff
@@ -306,7 +294,7 @@ These are the `UserSession.putFile`, `UserSession.getFile`, and `lookupProfile`
 
             </div>
           </div>
-        </div> : null
+        </div>
       );
     }
```
agraebe, line 299: we're implicitly asking to close the component here and remove the `componentWillMount` method. It confused me.

_browser/blockstack_storage.md
```diff
@@ -220,48 +220,36 @@ These are the `UserSession.putFile`, `UserSession.getFile`, and `lookupProfile`
...
 
     ```javascript
-    render() {
-      const { handleSignOut, userSession } = this.props;
-      const { person } = this.state;
-      const { username } = this.state;
+    export const Profile = ({ userData, handleSignOut }) => {
+      // ... state setup from before
 
       return (
```
agraebe, line 252: This is quite confusing as well. We're implicitly assuming the reader will know that the `render` method in the file is to be replaced by a `return` statement. Not sure if everyone would get that right away.

_browser/blockstack_storage.md
agraebe: Looking at the browser after this step, I get an error:
```
./src/App.js
Attempted import error: './Profile.js' does not contain a default export (imported as 'Profile').
```

_browser/blockstack_storage.md
```diff
@@ -315,61 +303,36 @@ These are the `UserSession.putFile`, `UserSession.getFile`, and `lookupProfile`
...
-        .then(() => {
-          this.setState({
-            statuses: statuses
-          })
-        })
+      await userSession.putFile('statuses.json', JSON.stringify(_statuses), options);
+      setStatuses(_statuses);
+      setStatusIndex(statusIndex + 1);
     }
     ```
```
agraebe, line 338: The `images/display-complete.png` doesn't seem to load

agraebe: Same error after saving the file: 

```
./src/App.js
Attempted import error: './Profile.js' does not contain a default export (imported as 'Profile').```

File: _browser/blockstack_storage.md [hunk 7]
```diff
@@ -396,48 +359,39 @@ Update `Profile.js` again.
...
+      setLoading(false);
     }
     ```
    By default, `getFile()` this method decrypts data; because the default `putFile()` encrypts it. In this case, the app shares statuses publicly. So, there is no need to decrypt.
 
-5. Call `fetchData()` from the `componentDidMount()` method. 
+5. Call `fetchData()` from the by using React's `useEffect` method, which will fetch data whenever the component's username state is changed.. 
 
     ```javascript
+    // after setting up your component's state
```
agraebe, line 390: might be not clear where this is in the file

_browser/blockstack_storage.md
```diff
@@ -578,8 +532,13 @@ process URL paths that contain the `.` (dot) character for example,
 3. Add a single method to the `Profile` class that determines if the app is viewing the local user's profile or another user's profile.
```
agraebe, line 532: The ` When you are done, the import statement should look like the following:` is in code style. Should remove the tabs

_browser/blockstack_storage.md
```diff
@@ -578,8 +532,13 @@ process URL paths that contain the `.` (dot) character for example,
...
 
     ```javascript
-    isLocal() {
-      return this.props.match.params.username ? false : true
+    // Make sure you add this new prop!
+    export const Profile = ({ userData, handleSignOut, match }) => {
+      // ...
+      const isLocal = () => {
```
agraebe, line 538: where in the file should this go? the "..." before and after make it difficult to understand that

_browser/blockstack_storage.md
```diff
@@ -630,40 +577,25 @@ process URL paths that contain the `.` (dot) character for example,
      documentation](http://blockstack.github.io/blockstack.js/#getfile) for
      details.
 
-5. Add the following block to `fetchData()` right after the call to `lookupProfile(username)... catch((error)=>{..}` block:
+5. Add the following block to `fetchData()` right after the call to `setUsername(username)` block:
```
agraebe, line 580: Why is this another step? Seems to add confusion and I don't understand the code better

_browser/blockstack_storage.md
```diff
@@ -630,40 +577,25 @@ process URL paths that contain the `.` (dot) character for example,
...
+    setStatuses(_statuses);
+    setLoading(false);
     ```
 
     This fetches the user statuses.
 
     Finally, you must conditionally render the logout button, status input textbox, and submit button so they don't show up when viewing another user's profile.
 
-6. Replace the `render()` method with the following:
+6. Replace the returned JSX in the `Profile` component with the following:
```
agraebe, line 595: The first time we reference "JSX". I'd rather avoid that

_browser/blockstack_storage.md
```diff
@@ -680,48 +612,46 @@ process URL paths that contain the `.` (dot) character for example,
...
             </div>
           </div>
         </div>
-      </div> : null
+      </div>
     );
-    }
     ```
 
     This checks to ensure that users are viewing their own profile, by wrapping the **Logout** button and inputs with the `{isLocal() && ...}` condition.
```
agraebe, line 657: At the end it says `Restart the application so that the disabling of the . (dot) rule takes effect.`... what is the dot rule?

_browser/blockstack_storage.md
aulneau: I would push back and say that just because it's not mentioned before doesn't mean we should not add it. The tutorial assumes some level of react experience.

_browser/blockstack_storage.md
hstove: This is very React-specific knowledge. TBH, I'm not sure how to put it. It's not a constructor/class anymore, it's a functional component, which is considered React best practice these days. You setup state using hooks in the beginning of the function.

_browser/blockstack_storage.md
agraebe: What's the win of adding "JSX" here as opposed to just referring to the `return` method? I don't see any upside but a potential confusion. 

Referring to JSX is fine by itself, I'm not saying we can't assume them to know about it. But, if we want to use the term, let's use it also in one of the first steps where we replace the render method with the return method?

_browser/blockstack_storage.md
hstove: I agree that it doesn't make much sense as a separate step.

_browser/blockstack_storage.md
hstove: I have literally no idea.

_browser/blockstack_storage.md
hstove: @aulneau there is some confusion because Alex was using the current version of the app generator, which uses a class component. But my updated PR for the generator uses functional components. So there is no `render` method.

hstove:
Looks like there are more images missing that were deleted in https://github.com/blockstack/docs.blockstack/pull/542 - I caught one of those and brought it back, but looks like there are more we need to fix.

hstove:
> There's also a link to the full sources that seem outdated: https://github.com/yknl/publik

This is ken's repo, because this tutorial is based on an app that he originally built. We should probably change this reference, and/or include it in our own Github somewhere.

_browser/blockstack_storage.md
aulneau: @hstove yeah def! I think it could be fine to rephrase it such that it references updating what is returned. I also think using JSX across the board is more specific and correct.

_browser/blockstack_storage.md
agraebe: I think my main confusion came from the issue that I used the old generator. It didn't "fit" with what I was looking at

_browser/blockstack_storage.md
timstackblock: > Modify the render() methodrendered result to add a text input and submit button to the by replacing it with the code below:  

This line needs to be updated

timstackblock:
Broken Image in line 107
 Network Connection

timstackblock:
Line 61 looks off

{% include scaffolding.md %}

hstove:
Hey @timstackblock and @agraebe , I've published the app generator updates to NPM, so you should be able to follow the tutorial, now, without any complicated developer steps, or having to run anything like the authenticator locally.

There is still other feedback like broken images and broken inline code that I do need to fix.

hstove:
I've updated the broken links and images.

timstackblock:
:shipit:

PR #586, opened 2020-05-07
feat: docs for handling redirect with connect

hstove:
Docs for handling authentication if the `postMessage` flow fails - which can happen on mobile.

### DIFF
--- a/_develop/connect/get-started.md
+++ b/_develop/connect/get-started.md
@@ -146,9 +146,9 @@ blockstackConnect.showBlockstackConnect(authOptions);
 
 ## Handling redirect fallbacks
 
-Connect is built to use popups with the `window.postMessage` API, which provides a much better and seamless user experience. However, there are times when this flow can fail. For example, the popup may be blocked, or the `window.postMessage` API might not work properly (which often happens on mobile browsers).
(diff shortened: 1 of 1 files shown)


_develop/connect/get-started.md
```diff
@@ -142,4 +142,24 @@ Then, you can use API methods under the `blockstackConnect` global variable:
 ```javascript
 const authOptions = { /** See docs above for options */ };
 blockstackConnect.showBlockstackConnect(authOptions);
+```
+
+## Handling redirect fallbacks
+
+Connect is built to use popups with the `window.postMessage` API, which provides a much better and seamless user experience. However, there are times when this flow can fail. For example, the popup may be blocked, or the `window.postMessage` API might not work properly (which often happens on mobile browsers).
```
agraebe, line 149: nit, might link to https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

_develop/connect/get-started.md
```diff
@@ -142,4 +142,24 @@ Then, you can use API methods under the `blockstackConnect` global variable:
...
+
+To make sure your app handles this gracefully, you'll need to handle the case where authentication is performed through vanilla redirects. With redirects, your users will be sent back to your app at a URL like:
```
agraebe, line 151: "vanilla redirects" is something I got hung up on a bit - is this for vanilla JS redirects?

hstove:
Good feedback Alex, thanks!

PR #589, opened 2020-05-11
Use correct path to install neon node

friedger:
This PR
* uses path `testnet/stacksnode` instead of `testnet` to install stacks-node


hstove:
Thanks for catching this friedger! We've fixed this as part of the PR to add mining support.

PR #607, opened 2020-05-29
feat: connect tx signing docs

hstove:
We now have the full Connect transaction signing flow for the three TX types:

- contract call
- contract deploy
- stx transfer

This PR updates the Connect docs for this stuff, and does some clean up / improvement on the existing docs.

### DIFF
--- a/_develop/connect/use-with-clarity.md
+++ b/_develop/connect/use-with-clarity.md
@@ -7,19 +7,12 @@ permalink: /:collection/:path.html
 
 With Connect, you can interact with the Stacks 2.0 blockchain. You can allow your users to send transactions and interact with smart contracts.
 
-* TOC
-{:toc}
-
-## How it works
-
-For your app's users to be able to execute a smart contract function, they need to sign and broadcast a transaction. It's important that users remain control of the private keys that sign these transactions. Connect provides an easy-to-use workflow that allows your users to securely sign transactions.
-
-Connect allows you to open the authenticator with parameters indicating the details of the transaction - like the smart contract address, function name, and specific arguments. Your users get the chance to see these details, and then sign and broadcast the transaction in a single click. Their transaction will be securely signed and broadcasted onto the Stacks blockchain. After this is done, a callback is fired to allow you to update your app.
-
-## Usage
-
 <div class="uk-card uk-card-default uk-card-body">
 <h5 class="uk-card-title">Transaction signing is still in progress</h5>
+<p>
+  To use these features, make sure you install the <code>testnet</code> tag the <code>@blockstack/connect</code> NPM package. You can do this by running <code>npm install --save @blockstack/connect@testnet</code>, or by specifying <code>testnet</code> as the version in
+  your package.json file.
+</p>
 <p>
   The Stacks 2.0 blockchain is still in testnet, and our web app integration is also still in beta. In order to use transaction signing in your application, you need to use the configuration `authOrigin` with `@blockstack/connect`.
 </p>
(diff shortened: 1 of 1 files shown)


_develop/connect/use-with-clarity.md
```diff
@@ -33,34 +26,78 @@ Connect allows you to open the authenticator with parameters indicating the deta
 </pre>
 </div>
 
-### `ContractCallOptions`
+* TOC
+{:toc}
+
+## How it works
+
+For your app's users to be able to execute a smart contract function, they need to sign and broadcast a transaction. It's important that users remain control of the private keys that sign these transactions. Connect provides an easy-to-use workflow that allows your users to securely sign transactions.
```
aulneau, line 34: > users remain control
needs "in"

_develop/connect/use-with-clarity.md
```diff
@@ -87,40 +124,104 @@ const functionArguments = [
...
+Here is the interface for the options you can provide to `openContractDeploy`:
+
+```ts
+interface ContractDeployOptions {
+  contractSource: string;
+  contractName: string;
+  authOrigin?: string;
+  appDetails: AuthOptions['appDetails'];
+  finished: (data: FinishedTxData) => void;
+}
```
aulneau, line 208: Naming conflict between the original options for the underlying function. Does it matter if one is `codeBody` and the other is `contractSource`?

```tsx
export interface ContractDeployOptions {
  contractName: string;
  codeBody: string;
  senderKey: string;
  fee?: BigNum;
  nonce?: BigNum;
  network?: StacksNetwork;
  anchorMode?: AnchorMode;
  postConditionMode?: PostConditionMode;
  postConditions?: PostCondition[];
}
```

Also, could we allow the types for `openContractDeploy` to be `Partial<ContractDeployOptions>` from the stacks-transactions lib?

_develop/connect/use-with-clarity.md
```diff
@@ -33,34 +26,78 @@ Connect allows you to open the authenticator with parameters indicating the deta
...
+    name: 'SuperApp',
+    icon: 'https://example.com/icon.png'
+  },
+  finished: (data) => {
+    console.log('TX ID:', data.txId);
+    console.log('Raw TX:', data.txRaw);
+  },
+};
+
+await openContractCall(opts);
```
agraebe, line 73: might be a newb question, but why do we await the function if there's a callback (finished) provided?

_develop/connect/use-with-clarity.md
```diff
@@ -33,34 +26,78 @@ Connect allows you to open the authenticator with parameters indicating the deta
...
+interface ContractCallOptions {
   contractAddress: string;
   functionName: string;
   contractName: string;
   functionArgs?: any[];
   authOrigin?: string;
-  userSession?: UserSession;
-  appDetails?: AuthOptions['appDetails'];
-  finished?: (data: FinishedTxData) => void;
+  appDetails: AuthOptions['appDetails'];
```
agraebe, line 85: is this in reference to some class? Should link to it because I wouldn't know what the format here should be without the context

_develop/connect/use-with-clarity.md
```diff
@@ -33,34 +26,78 @@ Connect allows you to open the authenticator with parameters indicating the deta
...
+  finished: (data: FinishedTxData) => void;
```
agraebe, line 86: what format is this? what properties does it have?

_develop/connect/use-with-clarity.md
agraebe: seems like it's just name & icon. I'd explicitly put these in here

_develop/connect/use-with-clarity.md
aulneau: typically to avoid [floating promises](https://palantir.github.io/tslint/rules/no-floating-promises/)

_develop/connect/use-with-clarity.md
```diff
@@ -87,40 +124,104 @@ const functionArguments = [
 
 If you're using Typescript, these Clarity types can be imported as `ContractCallArgumentType` from `@blockstack/connect`.
```
agraebe, line 125: maybe a link to the class/file would be good?

_develop/connect/use-with-clarity.md
```diff
@@ -33,34 +26,78 @@ Connect allows you to open the authenticator with parameters indicating the deta
...
+contractAddress | string | false | The Stacks address that published this contract
+contractName | string | false | The name that was used when publishing this contract
+functionName | string | false | The name of the function you're calling. This needs to be a [public function](/core/smart/clarityRef.html#define-public).
+functionArgs | array | false | The arguments you're calling the function with. You'll need to provide the Clarity type with each argument. See the below section for details. Defaults to `[]`.
+appDetails | object | false | A dictionary that includes `name` and `icon`
+finished | function | false | A callback that is fired when the transaction is signed and broadcasted. Your callback will receive an object back with a `txId` and a `txRaw`, both of which are strings.
+authOrigin | string | true | The location of the authenticator. This is only necessary when developing the authenticator locally, or when using beta features. Defaults to `"https://app.blockstack.org"`.
 
-#### Passing Clarity types with function arguments
+### Passing Clarity types with function arguments
```
agraebe, line 100: Could we link to this article where we mention principals? https://docs.blockstack.org/core/smart/principals.html

_develop/connect/use-with-clarity.md
aulneau: This is the typescript interface for the appDetails key. It's using `AuthOptions['appDetails'];` so there is less duplication. We should link to the place that the `AuthOptions` interface is defined.

_develop/connect/use-with-clarity.md
aulneau: This is the format for typing of a function. the `FinishedTxData` is an interface with this shape:
```tsx
interface FinishedTxData {
    txId: string;
    txRaw: string;
}
```

agraebe: thanks! I just noticed it's mentioned in the table - just didn't see it when I read it

_develop/connect/use-with-clarity.md
agraebe: makes sense. the code sample doesn't define a method and there's no return. will let you make the call if this is important to keep

_develop/connect/use-with-clarity.md
hstove: I ended up just putting the explicit object interface as Alex suggested - it may be less technically exact, but it likely reduces confusion.

_develop/connect/use-with-clarity.md
hstove: I updated the argument to be `codeBody` instead, and have [this issue](https://github.com/blockstack/ux/issues/408) to move to API parity with stacks-transactions.

_develop/connect/use-with-clarity.md
hstove: I like the suggestion, but right now we don't have a stable branch that this code lives on yet. Typescript developers will most likely have IDE autocomplete, so importing this enum will show them the options automatically.

PR #610, opened 2020-06-02
feat: update node docs for argon

hstove:
This PR does the following:

- Updates miner conf's `bootstrap_node` to use the stable domain `testnet-master.blockstack.org`, which shouldn't change during later testnet phases.
- Updates miner conf's `mode` and `peer_host` for argon.
- Updated any CLI commands to use `argon` instead of `neon`, i.e. `stacks-node argon`.
- Removed any reference to "Neon network", and replaced it with more generic language like "testnet network".

### DIFF
--- a/_core/smart/neon-node.md
+++ b/_core/smart/neon-node.md
@@ -3,10 +3,10 @@ layout: smart
 description: "Run a Stacks Testnet Node"
 permalink: /:collection/:path.html
 ---
-# Running a Neon Testnet Node
+# Running a Testnet Node
 {:.no_toc}
 
-"Neon" is phase 1 of the Stacks 2.0 testnet. In Neon, you can run a node and connect it to a public network. This guide will walk you through downloading and running your own node in the Neon network.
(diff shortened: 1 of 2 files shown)


diwakergupta:
Right now the miner config in this PR is using both `testnet-master` and `argon`. For simplicity, I'd suggest using `argon` consistently, and we can update to a version that uses `testnet-master` later once we've done more testing.

FWIW this is what I'm running with:

```toml
[node]
rpc_bind = "0.0.0.0:20443"
p2p_bind = "0.0.0.0:20444"
bootstrap_node = "048dd4f26101715853533dee005f0915375854fd5be73405f679c1917a5d4d16aaaf3c4c0d7a9c132a36b8c5fe1287f07dad8c910174d789eb24bdfb5ae26f5f27@argon.blockstack.org:20444"
seed = ...
miner = true

[burnchain]
chain = "bitcoin"
mode = "argon"
peer_host = "argon.blockstack.org"
rpc_port = 18443
peer_port = 18444
```

PR #617, opened 2020-06-17
Update clarityRef.json to latest from stacks-blockchain

kantai:
This updates the clarityRef.json file from blockstack/stacks-blockchain@8054315

Grabbed via:

```
docker run -it blockstack/stacks-blockchain:latest blockstack-core docgen > clarityRef.json
```


johndmulhausen:
LGTM :) Thanks @kantai !

Issue #608, opened 2020-06-01
Add guidance for checking for transaction status in app


markmhx:
For Blockstack Connect docs

agraebe:
This is covered in the docs backlog for mainnet cc @johndmulhausen

markmhx:
Are we tracking that in another issue and if so, mind linking here?

markmhx:
☝🏻 ping

agraebe:
From my understanding, this is captured in here: https://github.com/blockstack/docs.blockstack/issues/636.

@markmhx are you maybe proposing to use Connect directly to check the tx status?

markmhx:
For this particular issue, I'm thinking we just need some sort of guidance in the docs (presumably in ["Signing transactions"](https://docs.blockstack.org/develop/connect/use-with-clarity.html)) for developers to call an API with the transaction ID and get its status. #636 looks relevant to me so I'll add a comment there specifying this particular need.

I do believe additionally that we'll probably want to add transaction status support directly into Connect at some point. But we can tackle that in a different issue.

agraebe:
Let me add that quickly

PR #648, opened 2020-07-10
feat: check tx status

agraebe:
irt #608

### DIFF
--- a/develop/connect/use-with-clarity.md
+++ b/develop/connect/use-with-clarity.md
@@ -257,6 +257,7 @@ Following the steps above, you'll receive a transaction ID (`txId`) as part of t
 const response = await fetch(`https://sidecar.staging.blockstack.xyz/sidecar/v1/tx/${txId}`)
 const txData = await response.json();
 console.log(txData.status);
+```
 
 The API will respond with a JSON object that includes a `tx_status` field. Once the response returns `success`, the transactions was successfully processed.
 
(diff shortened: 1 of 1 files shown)


develop/connect/use-with-clarity.md
```diff
@@ -246,4 +246,18 @@ const MyComponent = () => {
...
+```
+
+## Checking transaction status
+You may want to check the status of a transaction after signing and broadcasting it. For instance, your app might want to enable certain components only _after_ a transaction was successfully processed on the blockchain.
+
+Following the steps above, you'll receive a transaction ID (`txId`) as part of the `finished` callback method once your user signed the transaction. Equipped with the ID, you can use the [Stacks Blockstack API](https://docs.blockstack.org/common/core_ref.html#stacks-blockchain-api) to request the status of the transaction.
+
+```bash
+curl 'https://sidecar.staging.blockstack.xyz/sidecar/v1/tx/<txId>'
+```
```
hstove, line 258:
```suggestion
```js
const response = await fetch(`https://sidecar.staging.blockstack.xyz/sidecar/v1/tx/${txId}`)
const txData = await response.json();
console.log(txData.status);
```
```

hstove: We should use JS instead of curl, because this page is entirely in the context of writing web apps.

markmhx:
@hstove how's this look to approve and merge with Alex's change?

hstove:
My suggested change was missing the final backticks (```) so if you look at the Markdown preview, it's broken. I think its a bug in the Github "suggested change" feature.

agraebe:
@hstove fixed, thanks! can you approve please?

PR #670, opened 2020-07-22
feat: next with status check

agraebe:
This adds a status checker above the left side navigation. Outstanding for this PR would be:

- [x] refactor based on best practices (e.g. remove isomorphic-unfetch) provided by @aulneau 
- [x] update the status checker endpoint to the known URL as soon as this [PR is merged](https://github.com/yknl/stacks-chain-status/pull/3) with help from @yknl

Screen Shot 2020-07-31 at 9 39 23 AM

### DIFF
--- a/src/components/side-nav.tsx
+++ b/src/components/side-nav.tsx
@@ -12,6 +12,7 @@ import { getCapsizeStyles } from '@components/mdx/typography';
 import { Text } from '@components/typography';
 import { css } from '@styled-system/css';
 import { SmartLink } from '@components/mdx';
+
 const Wrapper: React.FC<BoxProps & { containerProps?: BoxProps }> = ({
   width = `${SIDEBAR_WIDTH}px`,
   containerProps,
--- a/src/components/status-check.tsx
+++ b/src/components/status-check.tsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import { Box, Flex, space, themeColor, color } from '@blockstack/ui';
+import { border } from '@common/utils';
+import { Link } from '@components/mdx';
+import { Text } from '@components/typography';
+import { CheckmarkCircleIcon, ExclamationMarkCircleIcon, Spinner } from '@blockstack/ui';
+import useSWR from 'swr';
+
+const fetcher = url => fetch(url).then(r => r.json());
+
+export const StatusCheck: React.FC<any> = () => {
+  const STATUS_CHECKER_URL = 'https://status.test-blockstack.com';
+  const { data, error } = useSWR(`${STATUS_CHECKER_URL}/json`, fetcher);
+  let status = false;
+
+  if (data && data.hasOwnProperty('masterNodePings') && data.masterNodePings.length > 1) {
+    status = data.masterNodePings[0].value;
+  }
+
+  return (
+    <Link href={STATUS_CHECKER_URL} target="_blank" textDecoration="none">
+      <Box pt={space('base-loose')} px={space('base')} display={['none', 'none', 'block', 'block']}>
+        <Flex
+          border={border()}
+          borderRadius="6px"
+          px={space('base-tight')}
+          py={space('tight')}
+          align="center"
+          _hover={{ borderColor: themeColor('blue.400'), cursor: 'pointer' }}
+        >
+          <Box mr={space('tight')}>
+            {!data && !error && (
+              <Spinner color={themeColor('ink.600')} speed="1s" thickness="2px" size="sm" />
+            )}
+            {error && !status && (
+              <ExclamationMarkCircleIcon
+                color={themeColor('feedback.error')}
+                size="20px"
+                alignItems="center"
+                justifyContent="center"
+              />
+            )}
+            {data && status && (
+              <CheckmarkCircleIcon
+                color={themeColor('feedback.success')}
+                size="20px"
+                alignItems="center"
+                justifyContent="center"
+              />
+            )}
+          </Box>
+          <Text fontSize={'14px'} color={themeColor('ink.600')}>
+            {!data && `Stack 2.0 testnet status`}
+            {data && `Stacks 2.0 testnet is ${status ? 'online' : 'offline'}`}
+          </Text>
+        </Flex>
+      </Box>
+    </Link>
+  );
+};
+
+const SpinnerComponent = ({ color }: { color: string }) => (
+  <Spinner color={color} speed="1s" thickness="2px" size="sm" />
+);


src/components/status-check.tsx
```diff
@@ -0,0 +1,63 @@
...
+                color={themeColor('feedback.error')}
+                h="20px"
+                w="20px"
+                alignItems="center"
+                justifyContent="center"
+              />
+            )}
+          </Box>
+          <Text fontSize={'14px'} color={themeColor('ink.600')}>
+            Stacks 2.0 Network is {status ? 'online' : 'offline'}
```
hstove, line 55: Can we rename this to `Stacks 2.0 testnet`? Current phrasing makes it look like mainnet.

package.json
```diff
@@ -31,6 +31,7 @@
     "github-slugger": "^1.3.0",
     "gray-matter": "^4.0.2",
     "html-react-parser": "^0.13.0",
+    "isomorphic-unfetch": "^3.0.0",
```
aulneau, line 34: Next.js has fetch build in, [see here.](https://nextjs.org/blog/next-9-4#improved-built-in-fetch-support)

src/components/side-nav.tsx
```diff
@@ -10,6 +10,7 @@ import dynamic from 'next/dynamic';
 const SearchBox = dynamic(() => import('./search'));
 import { SIDEBAR_WIDTH } from '@common/constants';
 import { HEADER_HEIGHT } from '@components/header';
+import StatusCheck from './status-check';
```
aulneau, line 13: We typically prefer named exports and absolute paths:

`import { StatusCheck } from '@components/status-check';`

src/components/status-check.tsx
```diff
@@ -0,0 +1,63 @@
...
+  React.useEffect(() => {
+    fetch(`${STATUS_CHECKER_URL}/json`)
+      .then(r => r.json())
+      .then(data => {
+        if (data && data.masterNodePings.length > 1) {
+          setStatus(data.masterNodePings[0].value === 1);
+        }
+      })
+      .catch(err => console.log(err));
+  }, []);
```
aulneau, line 22: I would recommend using `useSWR` for this kind of thing: https://swr.now.sh/

src/components/status-check.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { Box, Flex, space, themeColor, color } from '@blockstack/ui';
+import { border } from '@common/utils';
+import { Link } from '@components/mdx';
+import { Text } from '@components/typography';
+import fetch from 'isomorphic-unfetch';
+import { CheckmarkCircleIcon, ExclamationMarkCircleIcon } from '@blockstack/ui';
+
+export const StatusCheck: React.FC<any> = () => {
+  const [status, setStatus] = React.useState(false);
+  const STATUS_CHECKER_URL = 'https://status-checker-blockstack.herokuapp.com';
```
hstove, line 11: Should this be `https://status.test-blockstack.com`?

src/components/status-check.tsx
aulneau: using `swr` will give us a lot nice things out of the box (from their site):
```
Fast page navigation
Polling on interval
Data dependency
Revalidation on focus
Revalidation on network recovery
```

src/components/status-check.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <Text fontSize={'14px'} color={themeColor('ink.600')}>
+            Stacks 2.0 Network is {status ? 'online' : 'offline'}
+          </Text>
+        </Flex>
+      </Box>
+    </Link>
+  );
+};
+
+export default StatusCheck;
```
kyranjamie, line 63:
```suggestion
```

src/components/side-nav.tsx
kyranjamie: Thomas beat me to it 
image

src/components/status-check.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  React.useEffect(() => {
+    fetch(`${STATUS_CHECKER_URL}/json`)
+      .then(r => r.json())
+      .then(data => {
+        if (data && data.masterNodePings.length > 1) {
+          setStatus(data.masterNodePings[0].value === 1);
+        }
+      })
+      .catch(err => console.log(err));
```
kyranjamie, line 21: Can we assume if the status check isn't working it's down? and `setStatus` here?

src/components/status-check.tsx
```diff
@@ -0,0 +1,63 @@
...
+          py={space('tight')}
+          align="center"
+          _hover={{ borderColor: themeColor('blue.400'), cursor: 'pointer' }}
+        >
+          <Box mr={space('tight')}>
+            {status ? (
+              <CheckmarkCircleIcon
+                color={themeColor('feedback.success')}
+                h="20px"
+                w="20px"
```
aulneau, line 40: these are causing a type failure, `size="20px"` should work

aulneau:
One other nice feature would be: since this is data fetching on the client, we should have a loading state for when it's initially loading data, so that it won't jar the flow of the page by showing up

src/components/status-check.tsx
aulneau: I think it's possible the status checker could be down while the testnet is available lol

package.json
agraebe: right, thanks!

src/components/status-check.tsx
agraebe: yes! with @yknl's update I can now use this URL instead :)

src/components/status-check.tsx
agraebe: @kyranjamie that makes sense to me

src/components/status-check.tsx
agraebe: nice!

src/components/status-check.tsx
agraebe: yes!

src/components/status-check.tsx
agraebe: changed that as well, thanks

agraebe:
> One other nice feature would be: since this is data fetching on the client, we should have a loading state for when it's initially loading data, so that it won't jar the flow of the page by showing up

Added that as well, thanks for the suggestion.

Now awaiting feedback from @jasperjansz and can finalize shortly.

jasperjansz:
Hey @agraebe and @aulneau , here are my thoughts:

It is (hopefully) an exception when the Testnet is down. It would be better to only show it when it's down rather than show the status at all times. Check this prototype: https://www.figma.com/proto/4XaD0Q2CBwuWCEkcMMAsfo/Docs?node-id=188%3A0&viewport=99%2C765%2C0.14496760070323944&scaling=min-zoom (try scrolling).

If we put it fixed at that position we can use this component on any page where testnet status is important.

agraebe:
@jasperjansz I really like that! @aulneau would you be able to carry this PR to the finish line as part of your bigger PR?

PR #692, opened 2020-08-06
Click to copy button

johndmulhausen:
Currently relies on `onlyText` in `utils`, however, that is returning code copies that have a lot of extra line breaks. Could use @aulneau 's advice!

### DIFF
--- a/src/components/mdx/components/code.tsx
+++ b/src/components/mdx/components/code.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { Box, BoxProps, color, space, themeColor, Highlighter, Button, useClipboard, utils } from '@blockstack/ui';
+import { Box, BoxProps, color, space, themeColor, Button, useClipboard } from '@blockstack/ui';
 import { border, onlyText } from '@common/utils';
 import { css } from '@styled-system/css';
 import { Text } from '@components/typography';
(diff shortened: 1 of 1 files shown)


src/components/mdx/components/code.tsx
```diff
@@ -39,6 +39,7 @@ export const Code: React.FC<
   BoxProps & { highlight?: string; lang?: string; lines: number }
 > = React.memo(
   React.forwardRef(({ children, highlight, lang, lines, ...rest }, ref) => {
+    const { hasCopied, onCopy } = useClipboard(onlyText(children));
```
aulneau, line 42: I would suggest this:
```tsx
// if you were to do `console.log(children)` you would see that it's an array of elements, each item in the array is a line.

console.log(children)

// this function takes a single child, uses the onlyText util, and then replaces all line breaks with an empty string (removes them)
const convertSingleChildToString = child => onlyText(child).replace(/\n/g, '');

// react children can be mapped over, but only when using the special helper from react, `import { Children } from 'react';` and then we use the above method to convert 
const tokenLines = Children.toArray(children).map(convertSingleChildToString);

console.log(tokenLines)

// now we will take the array of clean lines, and join them each with a single line break
const codeString = tokenLines.join('\n');

const { hasCopied, onCopy } = useClipboard(codeString);
```

johndmulhausen:
Pair programmed w/Thomas a bit to hash this out in `feat/next` directly so closing this PR

PR #693, opened 2020-08-06
fix: update stacks-blockchain-api url

zone117x:


aulneau:
Thanks for this @zone117x, I am going to close this and take care of it in #666

PR #723, opened 2020-08-11
Delete leading slash

JBotwina:
When calling the put, get, delete methods, refer to the file name without a leading slash.


JBotwina:
Glad to help!

PR #725, opened 2020-08-11
fix: imgix, github typo

aulneau:
### What this does

- This enables the use of imgix for production deployments (to serve up formats like webp when supported)
- Uses srcset to size images properly on mobile devices
- Fixes a typo in the github url
- I created a small rehype plugin to give us image sizes for local images. This gives us the ability to know the aspect ratio of our images and build it such that the page will never reflow when the image is loaded

### DIFF
--- a/src/components/feedback.tsx
+++ b/src/components/feedback.tsx
@@ -130,7 +130,7 @@ export const FeedbackSection: React.FC<BoxProps> = props => {
         mt={space(['extra-loose', 'extra-loose', 'base-loose'])}
       >
         <Link
-          href={`https://github.com/blockstack/docs/edit/master/src/${editPage}.md`}
+          href={`https://github.com/blockstack/docs/edit/master/src/pages${editPage}.md`}
           target="_blank"
           rel="nofollow noopener noreferrer"
           fontSize="14px"
--- a/src/components/mdx/image.tsx
+++ b/src/components/mdx/image.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Box, BoxProps } from '@blockstack/ui';
 import { useRouter } from 'next/router';
 
-const imgix = 'https://docs-stacks.imgix.net';
+const imgix = 'https://stacks-documentation.imgix.net';
 
 const params = '?auto=compress,format';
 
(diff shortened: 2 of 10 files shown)


src/components/feedback.tsx
```diff
@@ -130,7 +130,7 @@ export const FeedbackSection: React.FC<BoxProps> = props => {
         mt={space(['extra-loose', 'extra-loose', 'base-loose'])}
       >
         <Link
-          href={`https://github.com/blockstack/docs/edit/master/src/${editPage}.md`}
+          href={`https://github.com/blockstack/docs/edit/master/src/pages${editPage}.md`}
```
hstove, line 133: This doesn't need a trailing slash? The original looks like `editPage` wouldn't start with a slash.

src/components/mdx/md-contents.tsx
```diff
@@ -38,7 +38,7 @@ export const styleOverwrites = {
     pre: {
       // px: space(['none', 'none', 'extra-loose', 'extra-loose']),
     },
-    img: {
+    '.img': {
```
hstove, line 41: Why does it need the `.`?

src/components/mdx/image.tsx
```diff
@@ -18,27 +18,72 @@ const getUrl = pathname => {
...
+
+    const aspectRatio = getAspectRatio(dimensions);
+
+    // the content section will
+    const width = dimensions.width <= 720 ? dimensions.width : '100%';
+    return (
+      <Box width={width} maxWidth="100%" padding="0 !important" position="relative" className="img">
+        <Box height="0" paddingBottom={`${aspectRatio}%`} width="100%" />
+        <BaseImg position="absolute" top={0} left={0} {...props} />
+      </Box>
```
hstove, line 85: Slick :)

src/components/mdx/md-contents.tsx
aulneau: previously the spacing on images was by tag, but we want to treat the top level element as the spacing control, and with the aspect ratio changes, the top level element can sometimes be a `div`, so both have the className `img`.

src/components/feedback.tsx
aulneau: the pathname is always starts with `/`

PR #738, opened 2020-08-13
Fix possible bugs in claim-from-faucet

adamdunkels:
The `claim-from-faucet` function seems to have two minor bugs in it:
* the `map-set` function should take three arguments, map-name, tuple_A, and tuple_B, but it seems like it was just taking two arguments
* the `err` function seems like it should take a `u1` and not a `1`

### DIFF
--- a/src/pages/smart-contracts/principals.md
+++ b/src/pages/smart-contracts/principals.md
@@ -106,9 +106,9 @@ faucet" could be implemented as so:
   (if (is-none (map-get? claimed-before (tuple (sender tx-sender))))
       (let ((requester tx-sender)) ;; set a local variable requester = tx-sender
         (begin
-            (map-set claimed-before { sender: requester, claimed: true })
+            (map-set claimed-before { sender: requester } { claimed: true })
             (as-contract (stx-transfer? u1 tx-sender requester))))
-      (err 1)))
+      (err u1)))
 ```
 
 In this example, the public function `claim-from-faucet`:
--- a/src/pages/smart-contracts/principals.md
+++ b/src/pages/smart-contracts/principals.md
@@ -107,12 +107,11 @@ faucet" could be implemented as so:
  ((claimed bool)))
 
 (define-public (claim-from-faucet)
-  (if (is-none (map-get? claimed-before (tuple (sender tx-sender))))
-      (let ((requester tx-sender)) ;; set a local variable requester = tx-sender
-        (begin
-            (map-set claimed-before { sender: requester } { claimed: true })
-            (as-contract (stx-transfer? u1 tx-sender requester))))
-      (err u1)))
+  (if (is-none (map-get? claimed-before {sender: tx-sender}))
+    (let ((requester tx-sender))
+      (map-set claimed-before {sender: requester} {claimed: true})
+      (as-contract (stx-transfer? u1 tx-sender requester)))
+    (err u0)))
 ```
 
 In this example, the public function `claim-from-faucet`:
--- a/src/pages/smart-contracts/principals.md
+++ b/src/pages/smart-contracts/principals.md
@@ -111,11 +111,11 @@ faucet" could be implemented as so:
 (define-constant stx-amount u1)
 
 (define-public (claim-from-faucet)
-  (if (is-none (map-get? claimed-before {sender: tx-sender}))
-    (let ((requester tx-sender))
-      (map-set claimed-before {sender: requester} {claimed: true})
-      (as-contract (stx-transfer? u1 tx-sender requester)))
-    (err u0)))
+    (let ((requester tx-sender)) ;; set a local variable requester = tx-sender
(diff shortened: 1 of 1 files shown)


johndmulhausen:
@lgalabru - Everything look ok here?

hstove:
> the err function seems like it should take a u1 and not a 1

I don't think it matters - all that matters is that the function signature is consistent for any `err` response it can return. So, you could have `(err u1)`, or `(err 1)`, but you couldn't have both types of error responses in the same function.

psq:
that example could be seen as being a bit confusing, as it either should use the `(tuple ...)` syntax consistently, or the `{...}` syntax, but not both.  And for completeness' sake, it should also include the `map` definition...

njordhov:
The `let` special form allows multiple expressions in the body, making the `begin` redundant and unnecessary.

njordhov:
The [stx-transfer](https://docs.blockstack.org/references/clarity-language#stx-transfer) function returns `(err u1)` when the _sender does not have enough balance to transfer_. Should `claim-from-faucet` really return the same error value when there is an entry for the `tx-sender` in the `claimed-before` table? Consider changing the error value to say `u0`.

njordhov:
The indent is a bit off. Here is an alternative formatting:

```clojure 
(define-map claimed-before 
 ((sender principal))
 ((claimed bool)))

(define-public (claim-from-faucet)
  (if (is-none (map-get? claimed-before {sender: tx-sender}))
    (let ((requester tx-sender))
      (map-set claimed-before {sender: requester} {claimed: true})
      (as-contract (stx-transfer? u1 tx-sender requester)))
    (err u0)))
```

njordhov:
The example has a security risk as it doesn't verify that `map-set` succeeds. If setting `claimed-before` for the requester fails, they'd be able to make multiple faucet requests. 

Related: 
https://github.com/blockstack/stacks-blockchain/issues/1409
https://github.com/blockstack/stacks-blockchain/issues/972

njordhov:
The use of `as-contract` to get the _contract's principal_ complicates the example. What about Clarity having a symbol that provides the _contract's principal_ directly without the context switch?

adamdunkels:
As a noob trying to learn the language, the lack of a definition of the map certainly was a hurdle. (But it was a actually pretty good exercise for the reader to define it 😄 .)

I pushed the suggested changes by @njordhov to the branch.

psq:
> The example has a security risk as it doesn't verify that map-set succeeds

@njordhov how can `map-set` fail?  It either inserts or replaces the current entry.

njordhov:
> how can map-set fail?
It can for example fail if called in RPC read-only mode - see https://github.com/blockstack/stacks-blockchain/issues/1641
Although in this special case, this may not become an issue as `stx-transfer` will balk. My concern is more about that using this pattern generally to protect a resource may introduce potential exploits if a black hat figures out how to make `map-set` or other essential mutating call fail silently.

src/pages/smart-contracts/principals.md
```diff
@@ -102,13 +102,16 @@ For example, a smart contract that implements something like a "token
...
-  (if (is-none (map-get? claimed-before (tuple (sender tx-sender))))
-      (let ((requester tx-sender)) ;; set a local variable requester = tx-sender
-        (begin
-            (map-set claimed-before { sender: requester, claimed: true })
-            (as-contract (stx-transfer? u1 tx-sender requester))))
-      (err 1)))
+  (if (is-none (map-get? claimed-before {sender: tx-sender}))
+    (let ((requester tx-sender))
+      (map-set claimed-before {sender: requester} {claimed: true})
+      (as-contract (stx-transfer? u1 tx-sender requester)))
```
lgalabru, line 113: Thank you @adamdunkels, apologies for latency on this review!
Instead of `if` I think I would use the pattern "assert / mutate", and clearly explicit the fact that  an error occurring during `stx-transfer?` is being cascaded (it's the last statement), vs being shallowed.
```
(define-map claimed-before
  ((sender principal))
  ((claimed bool)))

(define-constant err-already-claimed u1)
(define-constant err-faucet-empty u2)

(define-public (claim-from-faucet)
    (let ((requester tx-sender) (amount u1)) ;; set a local variable requester = tx-sender
        (asserts! (is-none (map-get? claimed-before {sender: requester})) (err err-already-claimed))
        (unwrap! (as-contract (stx-transfer? amount tx-sender requester)) (err err-faucet-empty))
        (map-set claimed-before {sender: requester} {claimed: true})
        (ok amount)))
```

aulneau: @adamdunkels mind integrating these changes into your pr?

adamdunkels: @aulneau done!

aulneau: Wonderful. Do you think the list explaining the code below needs to be updated to reflect the new syntax changes?

PR #750, opened 2020-08-24
Simplify rust installation instructions

### LINKED ISSUE #747: Simplify rust installation instructions
Under the [Prerequisites](https://docs.blockstack.org/stacks-blockchain/testnet-node#prerequisites) section of the testnet-node setup, there is an easier way to install Rust with a single command.

The current procedure:

1. Run `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
2. Select option 1 to proceed (default)

To skip the second step and specify the default from the beginning, you could use a slight variation of the command:

1. Run `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y`

This is what I use [in my testnet shell script](https://github.com/AbsorbingChaos/bks-setup-miner/blob/master/config-miner-argon.sh#L76), and is further documented in the [rustup-init shell script](https://github.com/rust-lang/rustup/blob/master/rustup-init.sh#L28).

```
The installer for rustup

USAGE:
    rustup-init [FLAGS] [OPTIONS]

FLAGS:
    -v, --verbose           Enable verbose output
    -q, --quiet             Disable progress output
    -y                      Disable confirmation prompt.
        --no-modify-path    Don't configure the PATH environment variable
    -h, --help              Prints help information
    -V, --version           Prints version information
```

whoabuddy:
Fixes #747

When I first opened the markdown file in my editor it had extra line breaks that I do not see in the website, but rendered funny in my markdown preview, so I removed them as part of a217cc2. Otherwise this was a pretty straightforward update to the instructions.

### DIFF
--- a/src/pages/stacks-blockchain/testnet-node.md
+++ b/src/pages/stacks-blockchain/testnet-node.md
@@ -29,10 +29,6 @@ Ensure that you have Rust installed. If you are using macOS, Linux, or another U
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 ```
 
-If Rust is already installed, you might see this prompt. Select 'Proceed with Installation' to make sure you have the latest version installed.
-
-![rustup prompt](/images/rust-install.png)
-
 In case you just installed Rust, you will be prompted to run the following command to make the `cargo` command available:
 
 ```bash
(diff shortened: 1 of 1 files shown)


whoabuddy:
@aulneau or @agraebe I think this one is good to go!

PR #751, opened 2020-08-26
feat: add info about api client lib

agraebe:
Adding client lib info, related to this PR: https://github.com/blockstack/stacks-blockchain-api/pull/202

### DIFF
--- a/src/pages/references/stacks-blockchain.md
+++ b/src/pages/references/stacks-blockchain.md
@@ -5,9 +5,13 @@ description: Interacting with the Stacks 2.0 Blockchain via API
 
 ## Introduction
 
-The Stacks 2.0 Blockchain API was built to maintain pageable materialized views of the Stacks 2.0 Blockchain. It is a server that exposes a RESTful JSON API, hosted by PBC. It introduces aidditonal functionality (e.g. get all transactions), as well as proxies calls directly to Stacks Node. [You can find the OpenAPI specification and documentation here](https://blockstack.github.io/stacks-blockchain-api/).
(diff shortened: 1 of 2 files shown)


src/pages/references/stacks-blockchain.md
```diff
@@ -5,10 +5,14 @@ description: Interacting with the Stacks 2.0 Blockchain via API
 
 ## Introduction
 
-The Stacks 2.0 Blockchain API was built to maintain pageable materialized views of the Stacks 2.0 Blockchain. It is a server that exposes a RESTful JSON API, hosted by PBC. It introduces aidditonal functionality (e.g. get all transactions), as well as proxies calls directly to Stacks Node. [You can find the OpenAPI specification and documentation here](https://blockstack.github.io/stacks-blockchain-api/).
+The Stacks 2.0 Blockchain API was built to maintain pageable materialized views of the Stacks 2.0 Blockchain. It is a server that exposes a RESTful JSON API and WebSockets, hosted by PBC. It introduces aidditonal functionality (e.g. get all transactions), as well as proxies calls directly to Stacks Node. [You can find the OpenAPI specification and documentation here](https://blockstack.github.io/stacks-blockchain-api/).
```
zone117x, line 8: Curious if we want to emphasize the fact that we host an instance, encouraging devs to use ours rather than deploy their own? (which will be more straightforward to do so once https://github.com/blockstack/stacks-blockchain-api/pull/204 is merged)

agraebe: Yeah, we want to. Tl;dr is we already do in a way that should work IMHO. We show a very visible note on the page and point it out in another section below

Issue #633, opened 2020-07-02
Remove user-facing wallet docs


johndmulhausen:
- Nuke current Wallet stuff that is user-facing
- Focus remains on developers who are going to create their own wallets
- Integration docs will come from hardware wallet partners

markmhx:
I'm thinking all these pages can be removed:

- https://docs.blockstack.org/org/secureref.html
- https://docs.blockstack.org/org/tokenholders.html
- https://docs.blockstack.org/org/wallet-troubleshoot.html
- https://docs.blockstack.org/org/wallet-use.html
- https://docs.blockstack.org/org/wallet-install.html
- https://docs.blockstack.org/org/wallet-intro.html
- https://docs.blockstack.org/org/token.html

@jeffdomke that look right to you?

agraebe:
This seems to be mostly a task to delete existing content. If there's more context, please let me know.

PR #752, opened 2020-08-26
feat: remove user facing wallet docs

agraebe:
fixing #633


next.config.js
```diff
@@ -248,25 +248,11 @@ async function redirects() {
...
-    {
-      source: '/org/wallet-troubleshoot.html',
-      destination: '/stacks-wallet/troubleshooting',
-      permanent: true,
-    },
-    {
-      source: '/org/tokenholders.html',
-      destination: '/ecosystem/stacks-token-holders',
-      permanent: true,
-    },
```
aulneau, line 269: @agraebe we can't just remove redirects. We should always keep them, and redirect them to somewhere new. Ideally we'd avoid a 404 page. Can you please update this and have it redirect either to the homepage, or a page dedicated to the wallet?

agraebe: great catch! updating that now. thanks for reviewing the PRs @aulneau!

aulneau:
@agraebe I am going to close this as we went another direction, and I assume a different PR will take care of these docs.

Issue #637, opened 2020-07-02
Stacking guide


johndmulhausen:
Stacking guide (exec summary of SIP007)
https://github.com/blockstack/stacks-blockchain/blob/develop/sip/sip-007-stacking-consensus.md

agraebe:
relevant PR: https://github.com/blockstack/stacks-blockchain/pull/1790

kantai:
I put together something along the lines of a runbook for getting started with Stacking on the testnet:

https://gist.github.com/kantai/c261ca04114231f0f6a7ce34f0d2499b

Not sure where this documentation should end up landing (or if there's someone already working on this), but I figured it might be useful here.

agraebe:
@kantai thanks for sharing! I'll capture the steps in the #755 PR once I verified them. I'll use the stacks2.0 CLI though to be consistent within the docs

PR #755, opened 2020-08-26
feat: adding stacking guide

agraebe:
fixing #637


src/common/navigation.yaml
```diff
@@ -56,8 +57,9 @@ sections:
         sections:
           - title: Tutorials
             pages:
+              - path: /integrate-stacking
               - path: /install-api
-              - path: /installing-memcached
+              - path: /install-memcached
```
aulneau, line 62: We have to be very careful with renaming paths. We need to create a redirect anytime we do for a page that has been active.

agraebe: great point, adding a redirect!

src/pages/stacks-blockchain/stacking.md
```diff
@@ -1,50 +1,54 @@
...
+Stacks token holders do not automatically receive Stacking rewards. Instead, they must:
 
-### Integrating stacking
+- Commit to participation before a reward cycle begins
+- Hold ~94.000 Stacks tokens
+- Lock up Stacks tokens for a specified period
+- Set a Bitcoin address to receive rewards
 
-Ready to integrate stacking in your wallet or exchange?
+Token holders will have to use a wallet or exchange that supports participation in Stacking.
```
psq, line 37: this is true, but not limited to these 2.  Stacking is done by calling a contract, so anything able to call this contract would also qualify.

src/pages/stacks-blockchain/stacking.md
```diff
@@ -1,50 +1,54 @@
...
+- A reward cycle consists of two phases: prepare and reward
+- During the prepare phase, miners decide on an anchor block and a reward set. Mining any descendant forks of the anchor block requires transferring mining funds to the appropriate reward addresses. The reward set is the set of Bitcoin addresses which will receive funds in the reward cycle
+- Miners register as leader candidates for a future election by sending a key transaction that both burns cryptocurrency (proof-of-burn) and spends energy (proof-of-work). The transaction also registers the leader's preferred chain tip (must be a descendant of the anchor block) and commitment of funds to 5 addresses from the reward set
+- Token holders register for the next rewards cycle by broadcasting a signed message that locks up associated Stacks tokens for a protocol-specified lockup period, specifies a Bitcoin address to receive the funds, and votes on a Stacks chain tip
+- The leader that commits to the winning chain tip and the peers who also burn for that leader collectively share the reward, proportional to how much each one burned
```
psq, line 53: multiple leaders can commit to the same chain tip (hopefully, they all do, or this means a fork), it is the leader that wins the sortition, and whoever supported him that will get the reward.
That leader either transfers (or burns) btc, depending on several factors.

src/pages/stacks-blockchain/stacking.md
```diff
@@ -1,50 +1,54 @@
...
+
+-> Fun fact: The Stacking consensus algorithm is implemented as a smart contract, using [Clarity](/smart-contracts/overview).
+
+- Progression in Stacking consensus happens over reward cycles (with a fixed length). In each reward cycle, a set of Bitcoin addresses are iterated over
+- A reward cycle consists of two phases: prepare and reward
+- During the prepare phase, miners decide on an anchor block and a reward set. Mining any descendant forks of the anchor block requires transferring mining funds to the appropriate reward addresses. The reward set is the set of Bitcoin addresses which will receive funds in the reward cycle
+- Miners register as leader candidates for a future election by sending a key transaction that both burns cryptocurrency (proof-of-burn) and spends energy (proof-of-work). The transaction also registers the leader's preferred chain tip (must be a descendant of the anchor block) and commitment of funds to 5 addresses from the reward set
```
psq, line 51: the key transaction is not a burn.  That amount is recoverable after a commit used the registered key (or when a key can be reused multiple times, after a miner no longer wishes to use that key)

psq: the commit transaction is what registers the chaintip the miner is going to mine if it wins

src/pages/stacks-blockchain/stacking.md
```diff
@@ -1,50 +1,54 @@
 ---
 title: Understand Stacking
-description: Introduction to the consensus algorithm of the Stacks 2.0 blockchain
+description: Introduction to the rewarding mechanism of Proof-of-Transfer
```
psq, line 3: "reward" might be a better choice here.

src/pages/stacks-blockchain/stacking.md
agraebe: yeah, good point. at this point in the flow, the reader doesn't know yet about stacking details like that it is just a contract. let me see how to be clear but also not overwhelming

src/pages/stacks-blockchain/stacking.md
agraebe: thanks!

src/pages/stacks-blockchain/stacking.md
agraebe: how would you rephrase the above to cover that? I'd keep in mind that this is an intro to the topic, so rather stay high level and simplify where possible

src/pages/stacks-blockchain/stacking.md
agraebe: 👍

agraebe:
closing in favor of #805

Issue #614, opened 2020-06-09
Guide: Integrate Stacks


jeffdomke:
Suggesting we have a guide for those wishing to understand the basics of Stacks transactions with the likely aim of integrating with a wallet or exchange. Could be as extensive as: https://github.com/blockstack/stacks-transactions-js/blob/master/README.md with more basic info about Stacks transactions/formats. Or, could only cover the basics of transactions and send/receive. 

@yknl maybe you have some opinions on this. Someone who wanted to learn the basics + integrate – where would you send them today?

yknl:
We should create a full Stacks integration guide that covers everything from Stacks blockchain basics to transactions. This could live as a separate section in the docs.

johndmulhausen:
https://paper.dropbox.com/doc/wyvZRBd6DpfhVsIfRuB65

Turn this into a real doc

johndmulhausen:
Needs to include websocket endpoint details which we haven't touched yet

PR #757, opened 2020-08-29
feat: adding stacks integration guide

agraebe:
fixing #614


src/pages/stacks-blockchain/overview.md
```diff
@@ -89,7 +51,24 @@ An example of this is Stacking.
 
 ### Stacking
 
-Stackers are Stacks (STX) Token holders who provide a valuable service to the network by locking up their STX for a certain
+Stackers are Stacks (Stacks) Token holders who provide a valuable service to the network by locking up their Stacks for a certain
 period of time. As a reward, Stackers receive the Bitcoin that miners transfer as part of Proof of Transfer (PoX).
 
 -> Proof of Transfer and Stacking are in active development and are coming soon
+
+## Design specification
```
diwakergupta, line 76: [nit] "Design specification" as a title sounds odd, esp. as the first line says "describing the implementation". I would just say "Protocol Specification" instead.

src/pages/stacks-blockchain/overview.md
```diff
@@ -89,7 +51,24 @@ An example of this is Stacking.
...
+system-wide issue, and for documenting design decisions.
+
+-> See [SIP 000](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-000-stacks-improvement-proposal-process.md) for more details about the SIPs process.
+
+- ✅ [SIP 001:](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-001-burn-election.md) Burn Election
+- ✅ [SIP 002:](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-002-smart-contract-language.md) Clarity, a language for predictable smart contracts
+- ✅ [SIP 003:](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-003-peer-network.md) Peer Network
+- ✅ [SIP 004:](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-004-materialized-view.md) Cryptographic Committment to Materialized Views
+- ✅ [SIP 005:](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-005-blocks-and-transactions.md) Blocks, Transactions, and Accounts
+- [ ] SIP 006: Clarity Execution Cost Assessment _(Q2 2020)_
```
diwakergupta, line 90: [nit] you can actually link to SIP 006, 007 and 008 -- they're all written.

src/pages/stacks-blockchain/running-testnet-node.md
```diff
@@ -13,9 +13,15 @@ images:
 
 ## Introduction
 
-The Stacks 2.0 testnet is currently in development. As part of the testnet, you can run a node and connect it to a public network. This guide will walk you through downloading and running your own node in the testnet network.
+-> Note: The Stacks 2.0 testnet is currently in development. As part of the testnet, you can run a node and connect it to a public network.
 
-### Prerequisites
+This tutorial will walk you through the following steps:
+
+- Download and install the node
```
diwakergupta, line 20: [nit] "download and install the node software"

src/pages/stacks-blockchain/running-testnet-node.md
```diff
@@ -13,9 +13,15 @@ images:
...
+- Running the node
```
diwakergupta, line 21: [nit] Consistent tense: "Running" and "Mining" vs. "Download and install" (so if we want to be consistent with latter, we might say "Run the node", "Mine Stacks token")

src/pages/stacks-blockchain/sending-tokens.md
```diff
@@ -0,0 +1,207 @@
+---
+title: Sending tokens
+description: Learn how to transfer tokens
```
diwakergupta, line 3: This document is very NodeJS specific. We should call that out in the Introduction?

Separately, I wonder if it makes sense to either add some language agnostic description and/or link out to where people can get more information (e.g. the relevant SIPs)?

src/pages/stacks-blockchain/sending-tokens.md
```diff
@@ -0,0 +1,207 @@
...
+
+First install the stacks transactions library, a JavaScript [BigNum](https://en.wikipedia.org/wiki/Arbitrary-precision_arithmetic) implementation, and the API client for the [Stacks 2.0 Blockchain API](/references/stacks-blockchain):
+
+```shell
+npm install --save @blockstack/stacks-transactions@0.6.0 bn.js @stacks/blockchain-api-client
+```
+
+## Step 2: Specifying a sender
+
+In order to build and sign transactions, you will need a Stacks privat key. You can easily generate a new, random Stacks 2.0 sender key (see ["Generating an account" from the previous tutorial](http://localhost:3000/stacks-blockchain/managing-accounts#step-2-generating-an-account)).
```
diwakergupta, line 46: [typo] "privat"

diwakergupta: Also, broken link (uses localhost)

src/pages/stacks-blockchain/sending-tokens.md
```diff
@@ -0,0 +1,207 @@
...
+
+To generate a token transfer transaction, we will be using the `makeSTXTokenTransfer()` transaction builder function:
+
+```js
+const recipient = 'SP3FGQ8Z7JY9BWYZ5WM53E0M9NK7WHJF0691NZ159';
+
+// amount of Stacks tokens to send (in microstacks). 1,000,000 microstacks are worth 1 STX token
+const amount = new BN(1000000);
+
+// skip automatic fee estimation
```
diwakergupta, line 84: I found this confusing. If we're skipping automatic fee estimation, shouldn't we be setting some meaningful values for the fee (or is 0 a useful default)?

src/pages/stacks-blockchain/sending-tokens.md
```diff
@@ -0,0 +1,207 @@
...
+...
+
+const transaction = await makeSTXTokenTransfer(txOptions);
+```
+
+The generation method will need a few more pieces of information, as specified in the `txOptions` object:
+
+| Parameter   | Description                                                                                                                      | Optional |
+| ----------- | -------------------------------------------------------------------------------------------------------------------------------- | -------- |
+| `recipient` | The recipient Stacks address in c32check format                                                                                  | **No**   |
```
diwakergupta, line 114: [nit] I don't think we've described what `c32check` format means anywhere in this doc. Maybe link to https://www.npmjs.com/package/c32check?

src/pages/stacks-blockchain/sending-tokens.md
```diff
@@ -0,0 +1,207 @@
...
+| `memo`      | A memo string to attach additional information to the transaction. This data is limited to 33 bytes                              | Yes      |
+
+### Estimating fees
+
+If not specified, the transaction builder will automatically estimate the fee. Estimated fee rate is supplied by a Stacks node so network access is required. The fee is calculated based on the estimate fee rate and the size of the transaction in bytes.
+
+Another way to estimate the fee is to use the `estimateTransfer()` function after you have constructed a transaction:
+
+```js
+estimateTransfer(transaction);
```
diwakergupta, line 129: Is this a function with side-effects? That is, it updates the `fee` value within the `transaction`? If yes, probably out of scope for this PR, but 1) I'd generally argue against functions with side-effects -- they are error-prone. and 2) the name should reflect what the function does. It's not just estimating, it's actually estimating _and_ setting the fee.

cc/ @yknl

src/pages/stacks-blockchain/sending-tokens.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Step 4: Broadcasting transaction
+
+Next, we will broadcast the transaction to the Testnet using the `network` object we created earlier:
+
+```js
+const txId = await broadcastTransaction(transaction, testnet);
+```
+
+As soon as the `broadcastTransaction` is completed, a transaction ID is returned.
```
diwakergupta, line 152: Maybe worth re-iterating here that the fact you have a txId doesn't mean the transaction has completed executing.

src/pages/stacks-blockchain/testnet.md
```diff
@@ -0,0 +1,31 @@
...
+## Testnet phases
+
+- ✅ **Phase 1 (Neon):** is a developer local setup, mono-node, assembling SIP 001, SIP 002, SIP 004 and SIP 005. With this version, developers can not only run Stacks 2.0 on their development machines, but also write, execute, and test smart contracts.
+- ✅ **Phase 2 (Argon):** is the current version of our public testnet. This testnet includes SIP 003, and will be an open-membership public network, where participants will be able to validate and participate in mining testnet blocks.
+- [ ] **Phase 3 (Krypton):** is the upcoming version that we're anticipating to ship in Q3 2020. This version focus on the PoX basics (miner get rewards, Stackers get BTC distributions, etc).
+- [ ] **Phase 4 (Mainnet)** is the fully functional version, that we're intending to ship in Q4 2020.
+
+## Testnet roadmap
+
+✅ Released features&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<input type="checkbox" disabled /> Upcoming
```
diwakergupta: Rather than duplicate the table here, can we point to the "Testnet phases" section at https://www.blockstack.org/testnet? Or we just want to manually keep the two in sync? (updates are infrequent, so probably fine)

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,214 @@
+---
+title: Transaction generation
+description: Guide to Stacks 2.0 transactions
+---
+
+## Introduction
+
+Transactions are the fundamental unit of execution in the Stacks blockchain. Each transaction is originated from a standard account, and is retained in the Stacks blockchain history for eternity. This guide helps you understand Stacks 2.0 transactions.
```
diwakergupta, line 8: [nit] readers won't know what we mean by "standard account". Is the intention to link this back to the `accounts` page once that's populated?

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Lifecycle
+
+Transactions go through phases before being finally confirmed, and available for all, on the Stacks 2.0 network.
+
+![Transaction lifecycle](/images/tx-lifecycle.png)
+
+- **Generate**: Transactions are assembled according to the encoding specification.
+- **Verify and sign**: Transactions are verified to confirm if they are well-formed. Required signatures are filled in.
+- **Broadcast**: Transactions are send to a node.
```
diwakergupta, line 25: [typo] "send" -> "sent"

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,214 @@
...
+- **Register**: A miner receives transactions and adds them to the "mempool", a holding area for all the pending transactions.
```
diwakergupta, line 26: Feel free to link "mempool" to a crypto glossary, e.g. https://academy.binance.com/glossary/mempool

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,214 @@
...
+- **Process**: Miners review the mempool and select transactions for the next blocked to be mined. Depending on the transaction type, different actions can happen during this step. For example, post-conditions could be verified for a token transfer, new tokens could be minted, or an attemp to call an existing smart contract method could be made.
```
diwakergupta: [typo] "blocked" -> "block"
"attemp" -> "attempt"

FYI I don't think new Stacks tokens can be minted as a result of processing a transaction -- smart-contract defined tokens could be though. Might be worth clarifying that. New STX is minted via mining, but block rewards are not paid out immediately (miners have to wait 100 blocks to get rewards for a block they mined)

diwakergupta: I see now, you're considering "coinbase transactions" as well. Those are somewhat different in that you won't generate and broadcast them like you would other transactions.

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,214 @@
...
+| Poison Microblock | Punish leaders who intentionally equivocate about the microblocks they package                                    |
+
+-> A sample of each transaction type can be found in the [Stacks Blockchain API response definition for transactions](https://blockstack.github.io/stacks-blockchain-api/#operation/get_transaction_by_id).
+
+## Post-conditions
+
+Transaction post-conditions are a feature meant to limit the damage smart contract bugs can do in terms of destroying a user's assets whenever a contract is instantiated or a public method of an existing contract is executed. Post-conditions are meant to be added by the user (or by the user's wallet software) at the moment they sign a transaction.
+
+Post-conditions are intended to be used to force a transaction to abort if the transaction would cause a principal to send an asset in a way that is not to the user's liking. For example, a user may append a post-condition saying that upon successful execution, their account's STX balance should have decreased by no more than 1 STX. If this is not the case, then the transaction would abort and the account would only pay the transaction fee of processing it.
```
diwakergupta, line 48: Some duplication in this 2nd paragraph. I think they could be tightened into a single paragraph?

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Attributes
+
+Each transaction includes a field that describes zero or more post-conditions that must all be true when the transaction finishes running. A post-conditiona includes the following information:
```
diwakergupta, line 52: [typo] "post-conditiona"

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,214 @@
...
+
+| **Attribute** | **Description**                                                                           |
+| ------------- | ----------------------------------------------------------------------------------------- |
+| Principal     | Sender of the transaction, can be a Stacks address or a contract                          |
+| Asset name    | Asset to apply conditions to (could be Stacks, fungible, or non-fungible tokens)          |
+| Comparator    | Compare operation to be applied (could define "how much" or "whether or not")             |
```
diwakergupta, line 58: Should we link to the spec or include some concrete examples of comparators or post-conditions?

diwakergupta (commented):
First off, thanks for such a great start Alex!

Starting to work my way through the PR, will probably have to split up the feedback. Submitting a first round of comments now, more to come :)

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,462 @@
...
+| Authorization   | Type of authorization (`0x04` for standard, `0x05` for sponsored) and [spending conditions](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-005-blocks-and-transactions.md#transaction-authorization) |
+| Post-conditions | List of post-conditions, each including a [type ID and varible-length condition body](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-005-blocks-and-transactions.md#transaction-post-conditions-1)   |
+| Payload         | Transaction type and variable-length [payload](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-005-blocks-and-transactions.md#transaction-payloads-1)                                                 |
+
+## Construction
+
+The easies way to construct well-formed transactions is by [using the Stacks Transactions JS library](https://github.com/blockstack/stacks-transactions-js#post-conditions). You can construct the follow transaction types:
```
diwakergupta, line 90: [typo] "easies" -> easiest

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,462 @@
...
+| List             | `(list max-len entry-type)`           | List of maximum length max-len, with entries of type entry-type                                                                                                                     |
+| Response         | `(response ok-type err-type)`         | Object used by public functions to commit their changes or abort. May be returned or used by other functions as well, however, only public functions have the commit/abort behavior |
+| Optional         | `(optional some-type)`                | Option type for objects that can either be (some value) or none                                                                                                                     |
+| Buffer           | `(buff max-len)`                      | Byte buffer or maximum length max-len                                                                                                                                               |
```
diwakergupta, line 163: Did you mean "Byte buffer *with* maximum length `max-len`"?

Should we also add ascii and utf-8 string types to the table?

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,462 @@
...
+  ..., // other transaction options
+  postConditions: [standardSTXPostCondition]
+}
+
+const transaction = await makeContractCall(txOptions);
+```
+
+## Serialization
+
+A well-formed transaction construct is encoded in [Recursive Length Prefix ("RLP")](https://eth.wiki/en/fundamentals/rlp). RLP encoding results in a variable-sized byte array.
```
diwakergupta, line 253: Should we include pointers to reference implementation for serialization, as well as for signature generation (in the next section)?

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,462 @@
...
+- `limit`: the number of list items return per response (max is `200`)
+- `offset`: the number of elements to skip (starting from `0`)
+- `total`: the number of all available list items
+- `results`: the array of list items (length of array is between 0 and the set limit)
+
+Using the `limit` and `offset` properties, you can paginate through the entire list by increasing the offset by the limit until you reach the end of the list (as indicated by the `total` field).
+
+### Filter
+
+The current API design doesn't allow to set filters for queries. In case you are looking for a specific type of transaction, it is best to paginate through the list and iterate over the results of each response.
```
diwakergupta, line 351: Reading this a reader might naturally ask "why"? Any context we can add on why filters are not supported, or if they are planned in future?

diwakergupta (commented):
Few more minor comments.

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,487 @@
...
+-> The information on this page is based on a design proposal. You can find more conceptual details in this document: [SIP 005: Blocks, Transaction, Accounts](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-005-blocks-and-transactions.md).
+
+## Lifecycle
+
+Transactions go through phases before being finally confirmed, and available for all, on the Stacks 2.0 network.
+
+![Transaction lifecycle](/images/tx-lifecycle.png)
+
+- **Generate**: Transactions are assembled according to the encoding specification.
+- **Verify and sign**: Transactions are verified to confirm if they are well-formed. Required signatures are filled in.
```
yknl, line 24: While the transaction libs do try to ensure the transaction is valid. The verification step happens on the node once it's broadcasted.

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,487 @@
...
+
+## Types
+
+The Stacks 2.0 supports a set of different transaction types:
+
+| **Type**          | **Description**                                                                                                   |
+| ----------------- | ----------------------------------------------------------------------------------------------------------------- |
+| Coinbase          | The first transaction in a new block (an entity holding several transactions). Used to register for block rewards |
+| Token transfer    | Asset transfer from a sender to a recipient                                                                       |
+| Contract creation | Contract instantiation                                                                                            |
```
yknl, line 38: I think it makes more sense to use "Contract deploy" here.

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,487 @@
...
+| Contract call     | Contract call for a public, non read-only function                                                                |
+| Poison Microblock | Punish leaders who intentionally equivocate about the microblocks they package                                    |
+
+-> A sample of each transaction type can be found in the [Stacks Blockchain API response definition for transactions](https://blockstack.github.io/stacks-blockchain-api/#operation/get_transaction_by_id).
+
+## Post-conditions
+
+Transaction post-conditions are a feature meant to limit the damage smart contract bugs can do in terms of destroying a user's assets whenever a contract is instantiated or a public method of an existing contract is executed. Post-conditions are meant to be added by the user (or by the user's wallet software) at the moment they sign a transaction.
```
yknl, line 46: They also stop malicious contract developers.

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,487 @@
...
+
+The easies way to construct well-formed transactions is by [using the Stacks Transactions JS library](https://github.com/blockstack/stacks-transactions-js#post-conditions). You can construct the following transaction types:
+
+- Stacks token transfer
+- Smart contract deploy
+- Smart contract function call
+
+When constructing transactions, it is requireq to set the network the transaction is intended for. This can be either mainnet or testnet. At the moment of this writing, the only available option is the [testnet network](/stacks-blockchain/testnet).
+
+-> Transactions can be constructed and serialized offline. Internet access is only required to broadcast the transaction to the network.
```
yknl, line 98: One snag here is that you need to know the up-to-date nonce for the account. Otherwise the transaction might not be valid.

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,487 @@
...
+}
+
+const transaction = await makeContractCall(txOptions);
+```
+
+## Serialization
+
+A well-formed transaction construct is encoded in [Recursive Length Prefix ("RLP")](https://eth.wiki/en/fundamentals/rlp). RLP encoding results in a variable-sized byte array.
+
+In order to broadcast transactions to and between nodes on the network, RLP data is represented in hexadecial (also called the **raw format**).
```
yknl, line 259: *hexadecimal

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,487 @@
...
+
+To support an API-friendly and human-readable representation, transactions are serialized into a JSON format.
```
yknl, line 261: The transaction's are serialized into a hex string for transmission. Only the blockchain API server will convert it into JSON for API call responses.

src/pages/stacks-blockchain/overview.md
```diff
@@ -20,62 +20,24 @@ replica as all other peers.
...
-
-- ✅ **Helium** is a developer local setup, mono-node, assembling SIP 001, SIP 002, SIP 004 and SIP 005. With this version, developers can not only run Stacks 2.0 on their development machines, but also write, execute, and test smart contracts. See the instructions below for more details.
-- [ ] **Neon** is the upcoming version of our public testnet, that we're anticipating will ship in Q2 2020. This testnet will ship with SIP 003, and will be an open-membership public network, where participants will be able to validate and participate in mining testnet blocks.
-- [ ] **Mainnet** is the fully functional version, that we're intending to ship in Q3 2020.
-
-## Features
-
-✅ Live features&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<input type="checkbox" disabled /> Upcoming
+[@page-reference | grid-small]
+| /stacks-blockchain/managing-accounts, /stacks-blockchain/sending-tokens, /stacks-blockchain/running-testnet-node
```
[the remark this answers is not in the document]
agraebe: This is actually the same content wrapped in a tutorial format. I agree that this will likely be linked to from different places. I'll add a reference whenever I'll come across content that would benefit from it

src/pages/stacks-blockchain/overview.md
```diff
@@ -20,62 +20,24 @@ replica as all other peers.
...
-
-## Stacks 2.0 features
+## Capabilities
 
 ### Clarity Smart Contracts
 
-A Smart Contract is code running on the Stacks Blockchain that executes autonomously. When your Smart Contract has no
-room for errors, Clarity is here to help. Clarity is the safest language for writing Smart Contracts, optimized for
-predictability and security.
+A Smart Contract is code running on the Stacks Blockchain that executes autonomously. Clarity is the safest language for writing Smart Contracts, optimized for predictability and security.
```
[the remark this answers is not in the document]
agraebe: Good point, I updated this and linked to the glossary!

src/pages/stacks-blockchain/overview.md
```diff
@@ -20,62 +20,24 @@ replica as all other peers.
...
 
 [@page-reference | inline]
 | /smart-contracts/overview
 
-### STX Mining
+### Stacks Mining
 
 Anyone can be a Stacks Miner. There are no special hardware or software requirements, all you need is Bitcoin. Instead
```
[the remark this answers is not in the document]
agraebe: 👍

src/pages/stacks-blockchain/transactions.md
agraebe: Good point! Let me add a column for an example and link to the principal doc

src/pages/stacks-blockchain/overview.md
```diff
@@ -20,62 +20,24 @@ replica as all other peers.
...
-of spending energy, Stacks miners transfer Bitcoin to holders of Stacks Token (STX) to mine a block. This mechanism is
+of spending energy, Stacks miners transfer Bitcoin to holders of Stacks Token (Stacks) to mine a block. This mechanism is
 called Proof of Transfer (PoX).
 
 [@page-reference | inline]
```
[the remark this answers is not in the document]
agraebe: I believe the intention here was that there are two tutorials. one to get a node running and another one (building on top of the first) to run a miner.

What is the overlap we can get rid of?

src/pages/stacks-blockchain/overview.md
agraebe: I actually made this a table with *proposed* and *active* indicators. 

- SIP 6 is not proposed
- SIP 7 is proposed but not active
- SIP 8 - I couldn't find in [the sip folder](https://github.com/blockstack/stacks-blockchain/tree/develop/sip). any pointers?

src/pages/stacks-blockchain/sending-tokens.md
agraebe: good point!

src/pages/stacks-blockchain/sending-tokens.md
agraebe: Good points, will call out both!

src/pages/stacks-blockchain/sending-tokens.md
agraebe: I just checked the [source](https://github.com/blockstack/stacks-transactions-js/blob/master/src/builders.ts#L504) and it returns just the fee, no side-effects. The code example should be better, ie setting a fee var and update the tx

src/pages/stacks-blockchain/testnet.md
agraebe: I think you are right! I just found a way to link directly to the section on the testnet phase (https://www.blockstack.org/testnet#phases). Will just do that instead!

src/pages/stacks-blockchain/transactions.md
agraebe: Yes, I updated that as well

src/pages/stacks-blockchain/transactions.md
agraebe: @diwakergupta the SIP doesn't include the string types either: https://github.com/blockstack/stacks-blockchain/blob/develop/sip/sip-002-smart-contract-language.md#types

src/pages/stacks-blockchain/transactions.md
agraebe: Would be helpful. @yknl could you help with that? Just like the steps you added for signing?

src/pages/stacks-blockchain/transactions.md
agraebe: Checking with @zone117x to make a callout for this. thanks!

src/pages/stacks-blockchain/transactions.md
agraebe: thanks!

src/pages/stacks-blockchain/transactions.md
agraebe: Good point, reworded it a bit!

src/pages/stacks-blockchain/transactions.md
agraebe: update: there's actually API capability to filter - it is not yet documented. will fix and update this content!

src/pages/stacks-blockchain/transactions.md
agraebe: this was now added

src/pages/stacks-blockchain/overview.md
agraebe: thanks!

src/pages/stacks-blockchain/network.md
```diff
@@ -0,0 +1,179 @@
+---
+title: Network
+description: Guide to the Stacks 2.0 network
+---
+
+## Tokens
+
+Stacks ("STX") are the native tokens on the Stacks 2.0 blockchain. The smallest fraction of a STX is one micro-STX. 1.000.000 micro-STX make one STX.
```
kyranjamie, line 8:
```suggestion
Stacks ("STX") are the native tokens on the Stacks 2.0 blockchain. The smallest fraction of a STX is one micro-STX. 1,000,000 micro-STX make one STX.
```

The German in you coming out 🇩🇪

src/pages/stacks-blockchain/network.md
```diff
@@ -0,0 +1,179 @@
...
+
+STX amounts should be stored as integers (8 bytes long), and represent the amount of micro-STX. For display purposes, micro-STX are divided by 1.000.000 (decimal precision of 6).
```
kyranjamie, line 10:
```suggestion
STX amounts should be stored as integers (8 bytes long), and represent the amount of micro-STX. For display purposes, micro-STX are divided by 1,000,000 (decimal precision of 6).
```

src/pages/stacks-blockchain/network.md
```diff
@@ -0,0 +1,179 @@
...
+curl 'https://stacks-node-api-latest.argon.blockstack.xyz/v2/fees/transfer'
+```
+
+The API will respond with the fee rate (as integer):
+
+```js
+1;
+```
+
+[The Stacks Transactions JS library](https://github.com/blockstack/stacks-transactions-js) supports fee estimation for ...
```
kyranjamie, line 28:
```suggestion
[The Stacks Transactions JS library](https://github.com/blockstack/stacks-transactions-js) supports fee estimation for:
```

src/pages/stacks-blockchain/network.md
```diff
@@ -0,0 +1,179 @@
...
+
+- token transfers (`estimateTransfer`)
+- contract deploys (`estimateContractDeploy`)
+- non read-only contract calls (`estimateContractFunctionCall`)
+
+-> For an implementation using a different language than JavaScript, please review [this reference implementation](https://github.com/blockstack/stacks-transactions-js/blob/master/src/builders.ts#L97).
+
+## Nonces
+
+Every account carries a nonce property that indicates the number of transactions processed for the given account. Nonces are one-time codes and are by 1 with every transaction.
```
kyranjamie, line 38:
```suggestion
Every account carries a nonce property that indicates the number of transactions processed for the given account. Nonces are one-time codes and are incremented by 1 on every transaction.
```

src/pages/stacks-blockchain/sending-tokens.md
```diff
@@ -0,0 +1,219 @@
...
+tags:
+  - tutorial
+images:
+  large: /images/pages/testnet.svg
+  sm: /images/pages/testnet-sm.svg
+---
+
+## Introduction
+
+This tutorial will walk you through the following steps:
```
kyranjamie, line 16:
```suggestion
This tutorial walks you through the following steps:
```

src/includes/stacks-api-pagination.md
```diff
@@ -0,0 +1,10 @@
+### Pagination
+
+To make API responses more compact, lists returned by the API are paginated. For lists, the response body includes:
+
+- `limit`: the number of list items return per response (max is `200`)
```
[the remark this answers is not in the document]
agraebe: That's right, thanks for pointing that out

src/pages/stacks-blockchain/accounts.md
```diff
@@ -0,0 +1,175 @@
...
+
+When an account is created, a 24-phrase mnemonic is generated. This is often referred to as the **seed phrase**. The seed phrase provides access to Stacks 2.0 accounts.
+
+!> If the seed phrase is lost, access to the associated account cannot be restored. No person or organization, including Blockstack, can recover a lost seed phrase.
+
+The easiest way to generate a new Stacks 2.0 account is to use the [Stacks 2.0 CLI](https://github.com/blockstack/cli-blockstack/tree/feature/stacks-2.0-tx):
+
+```shell
+# install CLI globally
+sudo npm install -g "https://github.com/blockstack/cli-blockstack#feature/stacks-2.0-tx"
```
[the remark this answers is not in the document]
agraebe: need @yknl to confirm that. this was purposefully added due install issues in the past

src/pages/stacks-blockchain/accounts.md
```diff
@@ -0,0 +1,175 @@
...
+| `mnemonic`           | A 24-phrase seed phrase used to access the account, generated using [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) with 256 bits of entropy |
+| `keyInfo.address`    | Stacks address for the account                                                                                                                                       |
+| `keyInfo.privateKey` | Private key for the account. Required for [token transfers](/stacks-blockchain/transactions#stacks-token-transfer) and often referred to as `senderKey`              |
+| `keyInfo.index`      | Nonce for the account, starting at 0                                                                                                                                 |
+| `keyInfo.btcAddress` | Corresponding BTC address for the account. A construct from the previous blockchain (Stacks 1.0), not used at the moment.                                            |
+
+Note that the new account is instantiated implicitly and does not require internet access. There is no need to manually instantiate it on the Stacks 2.0 blockchain.
```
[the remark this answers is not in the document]
agraebe: I believe the question asked was: "do account need to be created on the blockchain explicitly?". without the context, it might be odd though. how would you reword it?

src/pages/stacks-blockchain/accounts.md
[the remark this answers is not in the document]
yknl: Yes, a properly installed npm shouldn't need sudo. Also agree with showing `--global`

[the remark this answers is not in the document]
agraebe: I remember it didn't work without sudo several times in tests and on my machine, but I can remove it. Also added the global

kyranjamie (approved):

src/pages/stacks-blockchain/accounts.md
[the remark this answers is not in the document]
muneeb-ali: Maybe: A new address/account automatically exists for each new private key. There is no need to manually instantiate an account on the Stacks 2.0 blockchain.

src/pages/stacks-blockchain/testnet.md
```diff
@@ -0,0 +1,24 @@
...
+title: Testnet
+description: Learn about testnet phases, timelines, and features
+images:
+  large: /images/pages/testnet.svg
+  sm: /images/pages/testnet-sm.svg
+---
+
+## Introduction
+
+The Stacks 2.0 testnet is an alterative blockchain, to be used for testing. It is intended for developer experimentation and not built for production-ready integrations. The testnet is frequently reset and Stacks tokens, used as fees for operations on the testnet, can be obtained for free using a [faucet](https://www.blockstack.org/testnet/faucet).
```
muneeb-ali, line 11: Maybe not say "alternative blockchain" but rather "Stacks 2.0 testnet is a separate (throwaway) blockchain, to be used for testing. 

Also, do we make clear that testnet Stacks are different from real Stacks.

Might be good to use "Stacks (STX)" somewhere to make that connection.

src/pages/stacks-blockchain/testnet.md
```diff
@@ -0,0 +1,24 @@
...
+
+With the completion of the Stacks 2.0 blockchain, a production-ready, stable version will be released as "mainnet". While work for the mainnet is in progress, the testnet reflects the most recent implementation of the Stacks 2.0 blockchain.
+
+## Release phases
+
+The work towards mainnet is split up into four sequential phases:
+
+- ✅ **Phase 1:** is a developer local setup, mono-node, assembling SIP 001, SIP 002, SIP 004 and SIP 005. With this version, developers can not only run Stacks 2.0 on their development machines, but also write, execute, and test smart contracts.
+- ✅ **Phase 2:** is the current version of our public testnet. This testnet includes SIP 003, and will be an open-membership public network, where participants will be able to validate and participate in mining testnet blocks.
+- [ ] **Phase 3:** is the upcoming version that we're anticipating to ship in Q3 2020. This version focus on the PoX basics (miner get rewards, Stackers get BTC distributions, etc).
```
muneeb-ali, line 21: See http://whenmainnet.co for some language on phases.

Also, I've seen the use of both 4 phases and 5 phases. Might be good to make that consistent.

src/pages/stacks-blockchain/overview.md
```diff
@@ -89,7 +68,26 @@ An example of this is Stacking.
...
+Stacks improvement proposals (SIPs) are aimed at describing the implementation of the Stacks blockchain, as well as
+proposing improvements. They should contain concise technical specifications of features or standards and the rationale
+behind it. SIPs are intended to be the primary medium for proposing new features, for collecting community input on a
+system-wide issue, and for documenting design decisions.
+
+-> See [SIP 000](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-000-stacks-improvement-proposal-process.md) for more details about the SIPs process.
+
+| **Proposal**                                                                                                   | **Description**                                     |
+| -------------------------------------------------------------------------------------------------------------- | --------------------------------------------------- |
+| [SIP 001](https://github.com/blockstack/stacks-blockchain/blob/master/sip/sip-001-burn-election.md)            | Burn Election                                       |
```
muneeb-ali, line 86: We might want to think about reducing the no. of SIPs linked here. Can always include a link that lists all SIPs.

SIP-007 for example is a lot more important than others. SIP-001 "burn election" is a bit less relevant and might cause confusion between proof of burn (which Stacks 2.0 no longer uses) and PoX.

psq: if there is no consensus (80%) on which block should be the anchor block for the upcoming reward cycle, burn will be used, so PoB is still relevant, and SIP-001 is still relevant for determining the sortition winner for each block.  Maybe that SIP should be renamed to reflect all this?

src/pages/stacks-blockchain/testnet.md
agraebe: good points, let me update!

src/pages/stacks-blockchain/accounts.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Creation
+
+An account is generated from a 24-word mnemonic phrase. This is often referred to as the **seed phrase**. The seed phrase provides access to Stacks 2.0 accounts.
+
+!> If the seed phrase is lost, access to the associated account cannot be restored. No person or organization, including Blockstack, can recover a lost seed phrase.
+
+The easiest way to generate a new Stacks 2.0 account is to use the [Stacks 2.0 CLI](https://github.com/blockstack/cli-blockstack/tree/feature/stacks-2.0-tx):
+
+```shell
```
aulneau, line 29: this should be `bash`

src/pages/stacks-blockchain/testnet.md
agraebe: I think the testnet page shows only 4 phases as well, right? should we stick to that?

agraebe: fyi, updated language based on suggestion

aulneau (changes requested):

src/pages/stacks-blockchain/overview.md
agraebe: We should keep in mind that this particular page is the first one folks would look at when they learn about the stacks 2.0 blockchain. I think the list feels overwhelming and likely not going to engage them (how to know which of these one should look at? and why?). Removing some or changing up the order might also look confusing for first-timers. I think the best here is to simply remove the table and add a callout that all SIPs can be found in the SIP folder, more a full, in-depth understanding of each proposal

src/pages/stacks-blockchain/accounts.md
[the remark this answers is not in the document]
agraebe: I like it! thanks!

src/pages/smart-contracts/hello-world-tutorial.md
```diff
@@ -182,7 +182,7 @@ The following set of commands will achieve the same goals as the above workflow.
 Install an early release of the new Blockstack CLI for Stacks 2.0.
 
 ```bash
-sudo npm install -g "https://github.com/blockstack/cli-blockstack#feature/stacks-2.0-tx"
+npm install --global "https://github.com/blockstack/cli-blockstack#feature/stacks-2.0-tx"
```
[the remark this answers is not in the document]
agraebe: I was actually thinking about that and I settled on `npm` installs only because:
- we'd want to be consistent with that everywhere (even in beginners tutorials/guides)
- if we refer to yarn, we need to add that to the requirements sections in tutorials

I would rather keep it simple and only use npm and list node as req. Folks who know yarn should also be able to take the npm command and do it with yarn. wdyt?

src/pages/stacks-blockchain/accounts.md
agraebe: nice, thanks!

src/pages/stacks-blockchain/sending-tokens.md
```diff
@@ -0,0 +1,219 @@
...
+  broadcastTransaction,
+  estimateTransfer,
+  getNonce,
+  StacksTestnet,
+  privateKeyToString,
+} = require('@blockstack/stacks-transactions');
+const { TransactionsApi, Configuration } = require('@stacks/blockchain-api-client');
+
+const apiConfig = new Configuration({
+  fetchApi: fetch,
```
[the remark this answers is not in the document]
agraebe: good point, let me add a note

src/pages/stacks-blockchain/network.md
```diff
@@ -0,0 +1,179 @@
...
+
+[The Stacks Transactions JS library](https://github.com/blockstack/stacks-transactions-js) supports fee estimation for:
+
+- token transfers (`estimateTransfer`)
+- contract deploys (`estimateContractDeploy`)
+- non read-only contract calls (`estimateContractFunctionCall`)
+
+-> For an implementation using a different language than JavaScript, please review [this reference implementation](https://github.com/blockstack/stacks-transactions-js/blob/master/src/builders.ts#L97).
+
+## Nonces
```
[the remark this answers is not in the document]
agraebe: yeah, great idea. thanks!

src/pages/stacks-blockchain/overview.md
```diff
@@ -89,7 +68,26 @@ An example of this is Stacking.
 
 ### Stacking
 
-Stackers are Stacks (STX) Token holders who provide a valuable service to the network by locking up their STX for a certain
-period of time. As a reward, Stackers receive the Bitcoin that miners transfer as part of Proof of Transfer (PoX).
+Stackers are Stacks (STX) Token holders who help secure the network by effectively acting as validators on the network, incentivizing honest behavior on the part of miners. Stackers have to lock up STX for a certain period of time. As a reward, Stackers receive the Bitcoin that miners transfer as part of Proof of Transfer (PoX).
```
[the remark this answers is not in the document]
agraebe: good observation, there is no consistency. I'll update to "Stacks (STX) token"

src/common/navigation.yaml
```diff
@@ -45,17 +45,24 @@ sections:
...
-          - path: /wire-format
-          - path: /atlas
-            pages:
-              - path: /overview
-              - path: /how-atlas-works
-              - path: /usage
+          - path: /testnet
+          - path: /accounts
+          - path: /transactions
+          - path: /network
```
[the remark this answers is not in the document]
agraebe: so glad you pointed that out, thanks!

src/pages/stacks-blockchain/network.md
```diff
@@ -0,0 +1,179 @@
...
+    "testnet": {
+        "target_block_time": 120
+    },
+    "mainnet": {
+        "target_block_time": 600
+    }
+}
+```
+
+-> The finality of the Stacks 2.0 network matches the bitcoin network finality.
```
jeffdomke, line 69: Can imagine this confusing. Might edit here and add it to the statement at the top of this section: 

"...bitcoin network. This allows transactions on Stacks to inherit the same finality and security of the Bitcoin blockchain." That might be a little too generalized but could make more sense.

jeffdomke (commented):
This is great. Two higher level things: 

Intro language around what is the Stacks blockchain needs revision

I'm curious if you also considered a quick reference for integrators of sorts? Across all these pages you do answer everything an integrator wants to know, but they also need to search around for that stuff vs. a single table that lists topics and quick explanations that point to the specific sections in these pages. 

The reason I mention this is because I assume the first thing an integrator wants is to quickly validate a bunch of key assumptions around if we are compatible with their setup. They are trying to mentally map the stuff that will be trivial vs. challenging.

src/pages/smart-contracts/hello-world-tutorial.md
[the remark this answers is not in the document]
agraebe: tl;dr - ok to move ahead without yarn for now. in the future, we might get a component to toggle so users can select what they want

src/pages/stacks-blockchain/testnet.md
```diff
@@ -0,0 +1,26 @@
...
+title: Testnet
+description: Learn about testnet phases, timelines, and features
+images:
+  large: /images/pages/testnet.svg
+  sm: /images/pages/testnet-sm.svg
+---
+
+## Introduction
+
+The Stacks 2.0 testnet is a separate (throwaway) blockchain, to be used for testing. It is intended for developer experimentation and not built for production-ready integrations. The testnet is frequently reset and Stacks (STX) tokens, used as fees for operations on the testnet, can be obtained for free using a [faucet](https://www.blockstack.org/testnet/faucet).
```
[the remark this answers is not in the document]
agraebe: removing `throwaway` to avoid the ocnfusion

src/pages/stacks-blockchain/accounts.md
```diff
@@ -0,0 +1,175 @@
...
+| -------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `mnemonic`           | A 24-word seed phrase used to access the account, generated using [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) with 256 bits of entropy |
+| `keyInfo.address`    | Stacks address for the account                                                                                                                                     |
+| `keyInfo.privateKey` | Private key for the account. Required for [token transfers](/stacks-blockchain/transactions#stacks-token-transfer) and often referred to as `senderKey`            |
+| `keyInfo.index`      | Nonce for the account, starting at 0                                                                                                                               |
+| `keyInfo.btcAddress` | Corresponding BTC address for the account. A construct from the previous blockchain (Stacks 1.0), not used at the moment.                                          |
```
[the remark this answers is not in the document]
agraebe: that's true. we only mention bitcoin in the overview on the bottom. I'll add an explanation

src/pages/stacks-blockchain/transactions.md
```diff
@@ -0,0 +1,523 @@
...
+| Contract deploy   | `smart_contract`    | Contract instantiation                                                                                                                                                                                |
+| Contract call     | `contract_call`     | Contract call for a public, non read-only function                                                                                                                                                    |
+| Poison Microblock | `poison_microblock` | Punish leaders who intentionally equivocate about the microblocks they package                                                                                                                        |
+
+-> The current [Naming service](/naming-services/overview) is unrelated to Stacks 2.0. There is no naming-specific transaction type. However, the naming service functionality will be ported to Stacks 2.0 in the future and might be implemented through smart contracts.
```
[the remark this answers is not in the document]
agraebe: good point, thanks!

src/pages/stacks-blockchain/accounts.md
```diff
@@ -0,0 +1,175 @@
...
+
+Note that a new account automatically exists for each new private key. There is no need to manually instantiate an account on the Stacks 2.0 blockchain.
+
+-> Addresses are created by generating the [RIPEMD-160 hash](https://en.wikipedia.org/wiki/RIPEMD#RIPEMD-160_hashes) of the [SHA256](https://en.bitcoinwiki.org/wiki/SHA-256) of the public key. BTC addresses are encoded with [Base58Check](https://en.bitcoin.it/wiki/Base58Check_encoding). For Stacks addresses, [c32check](https://github.com/blockstack/c32check) is used. Deriving an address from a public key can be done without internet access, for instance using the c32check `c32addressDecode` method.
```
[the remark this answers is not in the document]
agraebe: fixed with a note right before this content

src/pages/stacks-blockchain/network.md
agraebe: much better, thanks!

agraebe:
> This is great. Two higher level things:
> 
> Intro language around what is the Stacks blockchain needs revision

Thanks, I changed that! Took bits from @muneeb-ali's updated stacks-blockchain README.

> 
> I'm curious if you also considered a quick reference for integrators of sorts? Across all these pages you do answer everything an integrator wants to know, but they also need to search around for that stuff vs. a single table that lists topics and quick explanations that point to the specific sections in these pages.
> 
> The reason I mention this is because I assume the first thing an integrator wants is to quickly validate a bunch of key assumptions around if we are compatible with their setup. They are trying to mentally map the stuff that will be trivial vs. challenging.

Let me DM you to understand this a bit better :)

next.config.js
```diff
@@ -337,6 +337,31 @@ async function redirects() {
...
+    {
+      source: '/stacks-blockchain/testnet-node',
+      destination: '/stacks-blockchain/running-testnet-node',
+      permanent: true,
+    },
+    {
+      source: '/stacks-blockchain/testnet-node',
+      destination: '/stacks-blockchain/running-testnet-node',
+      permanent: true,
+    },
```
aulneau, line 349: duplicates

next.config.js
```diff
@@ -337,6 +337,31 @@ async function redirects() {
       permanent: true,
```
aulneau, line 337: @agraebe can you go through and ensure that the old redirects are updated to? such as https://github.com/blockstack/docs/blob/feat/stacks-integration-guide/next.config.js#L226-L228

aulneau (changes requested):

next.config.js
agraebe: thanks, fixed

next.config.js
agraebe: jup, done!

aulneau (approved):

PR #759, opened 2020-09-01
Update install-api.md

dantrevino:
Add checkout stacks-1.0 command.  Python install utilizes the legacy 1.0 branch, not the current master so the user must checkout the legacy branch before running `pip install`


aulneau:
Thanks @dantrevino ! appreciate it.

PR #762, opened 2020-09-02
Fix link to source code of AuthOptions

friedger:
This PR
* replace the current link to the source code of `AuthOptions` from the connect repo to the ux repo.
* updates the optional flags in the properties table of `AuthOptions`

### DIFF
--- a/src/pages/authentication/connect.md
+++ b/src/pages/authentication/connect.md
@@ -50,8 +50,8 @@ The exact interface you'll use [is defined as](https://github.com/blockstack/ux/
 
 ```typescript
 export interface AuthOptions {
-  redirectTo?: string;
-  finished?: (payload: FinishedData) => void;
+  redirectTo: string;
+  finished: (payload: FinishedData) => void;
   sendToSignIn?: boolean;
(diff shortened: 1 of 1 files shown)


agraebe (approved):
thanks @friedger. cc @hstove

agraebe (approved):
thanks @friedger!

hstove:
Thank you for fixing the link!

RE: the changes to AuthOptions. I know those fields are technically not required as per the Typescript definition, but I would lean more towards making them mandatory. You really need the `finished` and `redirectTo` options to fully implement authentication. My only hesitation to changing the TS definition is that it would be a breaking change. I definitely do not want our docs to indicate that these fields are optional, because you need them to implement auth.

hstove (approved):

hstove:
LGTM! Thanks for updating the PR @friedger .

hstove:
I'm unable to merge this. Someone with Vercel access can authorize the deploy, which should make all status checks pass. Or, someone with admin rights in this repo (maybe @agraebe ?) can merge.

aulneau:
Merged!

PR #778, opened 2020-09-15
Add reference from main entry of Blockstack connect to signing txs

friedger:
This PR
* adds a connection from the main entry page for Connect (e.g. from github) to the tx signing section with Connect.


agraebe (approved):
thanks!

PR #785, opened 2020-09-24
Update tutorials for krypton miners

lgalabru:


CharlieC3 (approved):
LGTM :shipit:

CharlieC3:
We might need to change the sidecar URLs to remove the `-latest` part if we're going to get rid of one of the sidecar deployments. Will need @zone117x to confirm if the two sidecar deployments is still needed.

zone117x:
Two deployments is not needed. The URLs should be updated 👍

agraebe (approved):

wileyj:
I think we can further simplify the DNS names here

CharlieC3 (approved):
@wileyj Good catch on the `xyz` -> `org`!
:shipit: LGTM

lgalabru:
Thank you @wileyj!

PR #788, opened 2020-09-24
API documentation from stacks-blockchain

agraebe:
PR replacing #786 (to pass tests)


agraebe:
Will try to merge #790 instead

Issue #812, opened 2020-10-07
Stacking contract reference


agraebe:
The Stacking guide includes a [contract reference section](https://pr-805.docs.stacks.engineering/stacks-blockchain/stacking#stacking-contract). This section should be ...

- [x] Moved to the reference section in the docs
- [x] Auto-generated by upstream changes from `stacks-blockchain`

@lgalabru @kantai - do we need to create a new issue in the node project or is this good enough for you to prioritize this work?

cc @aulneau

kantai:
This issue works for me. I have a working implementation of the docs generation, and will open a PR for it in the next day or so, and will update this issue when I do.

kantai:
This PR adds generation of a JSON reference for boot contracts:

https://github.com/blockstack/stacks-blockchain/pull/1957

It also updates everyone's favorite PR robot so that it:

1. Formats the JSON using jq to make the PRs a little friendlier
2. Uses an e-mail address that should fix the issues with the CLA bot
3. Stores the new contracts reference JSON in the same directory as the Clarity reference.

agraebe:
awesome! @aulneau what do we need to do to on the docs side?

PR #815, opened 2020-10-09
feat: add stacking reference page

agraebe:
fixing #812

in reference to: https://github.com/blockstack/stacks-blockchain/pull/1957/

### DIFF
--- a/src/common/data/stacking-ref.ts
+++ b/src/common/data/stacking-ref.ts
@@ -1,0 +1,118 @@
+import { renderMdx } from '@common/data/mdx';
+import STACKING_REFERENCE from '../../_data/boot-contracts-reference.json';
+
+const wrapInClarityTicks = (string: string) => {
+  let newString = '```clarity';
+  newString += `
+`;
+  newString += string.trim();
+  newString += `
+`;
+  newString += '```';
+  return newString;
+};
+
+const inlineCode = (string: string) => {
+  let newString = '`';
+  newString += string.trim();
+  newString += '`';
+
+  if (newString === '``') {
+    newString = '';
+  }
+
+  return newString;
+};
+
+const generateMarkdown = () => {
+  let publicFunctions = '';
+  let readonlyFunctions = '';
+  let errorCodes = '';
+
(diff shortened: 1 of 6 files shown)


kantai (approved):
Looks great to me, thanks @agraebe !

src/pages/references/stacking-contract.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Stacking Contract
+description: See a detailed list of all functions and error codes of the Stacking contract.
+---
+
+export { convertStackingRefToMdx as getStaticProps } from '@common/data/stacking-ref'
+import { StackingErrorcodeReference, StackingFunctionReference } from '@components/stacking-ref'
+
+Stacking is implemented as a smart contract using Clarity. On the testnet, you can always find the Stacking contract by this contract identifier: `ST000000000000000000002AMW42H.pox`.
```
aulneau, line 9: We need to add some kind of h2 here so that this paragraph gets wrapped in a section, and as a result gets styled correctly. `Introduction` or `Overview` would work.

src/common/data/stacking-ref.ts
```diff
@@ -0,0 +1,118 @@
...
+    publicFunctions += `
+### ${entry.name}
+
+**Signature:** ${inlineCode(entry.signature)}
+
+
+**Input:** ${inlineCode(entry.input_type)}
+
+
+**Output:** ${inlineCode(entry.output_type)}
```
aulneau, line 42: we should try `wrapInClarityTicks` here so that we get formatting. I'd give it a try else where, too, so we can format it. Some of these outputs are very long and would benefit from the formatting.

aulneau (changes requested):

src/common/data/stacking-ref.ts
agraebe: I just gave it a try but it doesn't seem to improve UX. The way the input/output/signatures are formatted doesn't line up with clarity formatting rules, so the only benefit is that brackets are in a different color than the keys. The scrolling on one line could also make it challenging. I'd suggest we punt that for this PR and take note that we should redesign the reference page

aulneau (approved):

PR #820, opened 2020-10-14
[wip] docs: stacking guide updates for most recent pox implementation

agraebe:


agraebe:
@yknl is working on a stacking library that would invalidate this work. the new tutorial would be much simpler. closing in favor of the new lib docs updates in the future

Issue #639, opened 2020-07-02
Guide to local Clarity development


johndmulhausen:
Should include:

- [x] Setting up all components
- [x] Pointing all settings towards mocknet
- [x] Troubleshooting issues when developing contracts
- [x] Configuring mocknet setting (.e.g target block time)

PR #827, opened 2020-10-16
docs: add local environment setup instructions

agraebe:
fixing #639

### DIFF
--- a/src/pages/write-smart-contracts/setting-up-local-environment.md
+++ b/src/pages/write-smart-contracts/setting-up-local-environment.md
@@ -1,0 +1,206 @@
+---
+title: Setting up local environment
+description: Learn how to set up your local smart contract development environment
+duration: 30 minutes
+experience: intermediate
+---
+
+## Introduction
+
+As soon as you get to work on your smart contracts and testing them end to end within your application, you need a local development environment that supports all steps and helps you achieve faster iteration cycles. This is critical because transaction times on the testnet take a few minutes and would slow you down. The solution is to run a local version of the testnet, often called a "mocknet".
+
+By the end of this tutorial, you will:
+
+- Have all required tooling and services up and running locally
+- Be able to deploy and test smart contracts inside your app
+
+## Prerequisites
+
+### Prior tutorials
+
+You need to be familiar with Clarity and the Stacks authentication (using Connect). Please finish all previous tutorials, especially the [Counter tutorial](/smart-contracts/counter-tutorial) and the [Building a todo app tutorial](/authentication/building-todo-app).
+
+### Environment
+
+You should have [NodeJS](https://nodejs.org/en/download/) and [Docker](https://docs.docker.com/get-docker/) installed on your workstation.
+
+## Overview
+
+For a complete local mocknet setup, you need the following services:
+
+- [Stacks Node](https://github.com/blockstack/stacks-blockchain)
+- [Stacks Blockchain API](https://github.com/blockstack/stacks-blockchain-api)
+- [Stacks Explorer](https://github.com/blockstack/explorer)
+
+## Step 1: Spin up mocknet
+
+The easiest way to spin up the Node, Blockchain API, and the Explorer locally is with Docker:
+
+```bash
+git clone https://github.com/blockstack/stacks-local-dev
+cd stacks-local-dev/
+docker-compose up -d
+```
+
+-> Make sure your Docker service is up and running
+
+The initial start will take a few minutes. As soon as it is complete, you should be able to open up [http://localhost:3999/](http://localhost:3999/) and see:
+
+```js
+{"status":"ready"}
+```
+
+~> The Stacks Node is running on port `20443`, the Stacks Blockchain API on `3999` and the Stacks Explorer is on port `3000`.
+
+## Step 2: Deploy your contract
+
+~> Make sure you completed the [Counter tutorial](/smart-contracts/counter-tutorial). A working contract and test-driven development setup is recommended
+
+When you develop your contract in a test-driven fashion, you will aim to complete the contract before deploying it to the testnet. Only once the contract is ready to be tested inside your app (or together with other contracts), you will need to deploy it. The fastest way to do so is by using the CLI:
+
+```shell
+# install the new CLI if not done yet
+npm install --global @stacks/cli
+
+# deploy the contract:
+# deploy_contract <SOURCE_FILE> <CONTRACT_NAME> <FEE> <NONCE> <PAYMENT_KEY> -t -T <NETWORK_ADDRESS>
+blockstack deploy_contract ./counter.clar counter 2000 0 cb3df38053d132895220b9ce471f6b676db5b9bf0b4adefb55f2118ece2478df01 -t -T http://localhost:20443
+```
+
+-> You can use one of the [default testnet keys](https://github.com/blockstack/stacks-blockchain-api/blob/e6e445b2cd1055f4bf25af7af646405783d5877e/src/api/routes/debug.ts#L36-L52) as your payment key
+
+The CLI will respond with a new contract deploy transaction:
+
+```js
+{
+  txid: '4bc11e1334c52852111142d55e5bbcfadbbb3441c44aa29a4bc507f7d2239377',
+  transaction: 'https://testnet-explorer.now.sh/txid/0x4bc11e1334c52852111142d55e5bbcfadbbb3441c44aa29a4bc507f7d2239377'
+}
+```
+
+If you have to deploy a new version of your contract, it is much simpler to reset the local mocknet that to deploy a new contract with a different identifier. Changing contract identifiers across projects can cause mistakes. To reset the mocknet, you can run the following inside the `stacks-local-dev` folder:
+
+```bash
+# make sure you are inside the stacks-local-dev folder
+# shutdown and start back up
+docker-compose down; docker-compose up -d
+```
+
+## Step 3: Create an app
+
+With all the previous services running and a working smart contract, you are ready to test the contract inside an app. We will reuse the [app from the todo tutorial](/authentication/building-todo-app).
+
+-> If you already have this project, or your own app, you can skip this step
+
+```bash
+git clone https://github.com/blockstack/blockstack-todos.git
+cd blockstack-todos
+npm install --saveDev @stacks/network@1.0.0-beta.8
+npm install
+npm run start
+```
+
+The command will identify that port 3000 being used already (by the Explorer running inside Docker). Hit `Y` to accept using port 3001. Once the app is running, a new windoe will open up ([http://localhost:3001/](http://localhost:3001/)), displaying the homepage.
+
+### Add contract calls
+
+Let's add a few sample contract calls to the application! You will use the `counter.clar` contract that should be deployed already.
+
+Open the todo project with the editor of your choice. Inside the `src/components/TodoList.jsx` file, add a few more imports:
+
+```js
+import { openContractCall } from '@blockstack/connect';
+import { Button } from '@blockstack/ui';
+import { callReadOnlyFunction, cvToString } from '@blockstack/stacks-transactions';
+import { StacksMocknet } from '@stacks/network';
+```
+
+Next, add a read-only contract call method.
+
+-> Read-only contract calls do not require transactions to be created or broadcasted to the network. The read-only calls are regular API calls that return the result
+
+```js
+...
+// const { userSession } = authOptions;
+
+// load user data from session
+const userData = userSession.loadUserData();
+
+const doReadOnlyCall = async name => {
+  const opts = {
+    contractAddress: 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6',
+    contractName: 'counter',
+    functionName: name,
+    functionArgs: [],
+    network: new StacksMocknet(),
+    senderAddress: userData.profile.stxAddress,
+  };
+
+  const resp = await callReadOnlyFunction(opts);
+
+  console.log(cvToString(resp));
+};
+```
+
+Now, add a regular contract call method right below the privous code. This method will create and broadcast a new transaction to the network.
+
+```js
+const doContractCall = async name => {
+  const opts = {
+    contractAddress: 'STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6',
+    contractName: 'counter',
+    functionName: name,
+    functionArgs: [],
+    network: new StacksMocknet(),
+    appDetails: {
+      name: "To-do's",
+      icon: window.location.origin + '/logo.svg',
+    },
+    finished: data => {
+      console.log('TX ID:', data.txId);
+      console.log('Raw TX:', data.txRaw);
+    },
+  };
+
+  await openContractCall(opts);
+};
+```
+
+Finally, you need to add three buttons for the contract calls inside the `return` method on the bottom of the file:
+
+```js
+<Box mt={[5, '60px']}>
+  <Button onClick={() => doContractCall('increment')}>Increment</Button>
+  <Button onClick={() => doContractCall('decrement')}>Decrement</Button>
+  <Button onClick={() => doReadOnlyCall('get-counter')}>Get counter</Button>
+</Box>
+```
+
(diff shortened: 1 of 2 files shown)


src/pages/smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,332 @@
...
+```js
+{"status":"ready"}
+```
+
+## Step 2: Set up Stacks Explorer
+
+The Stacks Explorer makes it easy way to review transactions. Let's spin up a local version:
+
+```bash
+git clone https://github.com/blockstack/explorer.git
```
agraebe, line 57: @wileyj could we also just use a docker image for this project?

src/pages/smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,332 @@
...
+
+## Step 5: Set up Stacks Authenticator
+
+Users of your web app will need to authenticate using their Secret Key before they could sign a Stacks transaction. The authentication needs to point to your local mocknet.
+
+Let's spin up the authentication service:
+
+```bash
+cd ..
+git clone https://github.com/blockstack/ux.git
```
agraebe, line 178: @wileyj here's another thing we could discuss: maybe this could be another docker image that we pull into the compose file for this tutorial?

agraebe:
@yknl could you help me identify the changes needed after monorepo work is done?

- [ ] use `StacksMocknet`
- [ ] change dependencies

src/pages/smart-contracts/setting-up-local-environment.md
wileyj: yes! we have an image that we use for the prod explorer. working on a fully-dockerized instruction set. 
the image is here: `quay.io/blockstack/explorer:latest`

src/pages/smart-contracts/setting-up-local-environment.md
wileyj: For this I'll have to take a closer look, i'm not familiar with what's going on in this repo

wileyj: update: totally doable. i'll have to submit a PR to get it ready but we can adjust instructions later once that is ready as part of the full-docker setup

hstove:
At a high level, we should use the testnet demo app for a "template" of clarity web app development. This way, you don't have to fiddle with React code. https://github.com/blockstack/ux/tree/master/packages/test-app

src/pages/smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,332 @@
...
+const { StacksTestnet } = require('@stacks/network');
+
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+// set to local mocknet
+const network = new StacksTestnet();
+network.coreApiUrl = 'http://localhost:3999';
```
yknl, line 121: This can be changed to `const network = new StacksMocknet();` as soon as stacks.js monorepo is released and dependency updated.

friedger:
In my practise it is much easier to restart mocknet and redeploy the contract than finding the code for the contract addresses and replace them. In particular if you have more than one. 

The `yarn lerna.. ` command for the authenticator should be simplified to something like `yarn start-authenticator`

Could the explorer be part of stacks api docker image? Or have a mocknet docker image?

agraebe:
@dantrevino and @psq - any input for this? what is your local contract setup like?

dantrevino:
@agraebe ty for reaching out.  Instructions work and seem clear enough, however ... I feel like this is useful for either a) those getting started from scratch, and/or b) bigger teams that have a real distinction between their QA and dev teams.  Either that or I personally lack a good process, which is highly likely.

One nitpick ... align on a single node package manager. Instructions are mixed between npm and yarn.

None of this precludes the PRs usefulness.

For me personally, I have the following issues:
1) Setting up my own authenticator means I have to track a completely different set of different user accounts in a different set of browsers.  Probably good practice, but not very likely for lone developers.  I.e. I see a lot of value in testing with my real-world (test) accounts that I use across multiple apps, including apps that are not mine.  This could become even more required if Collections ever becomes reality.
2) Instructions for doing an `npm init clarity-starter` within an **existing** project would likely be easier for devs who want to bring Clarity to their existing applications.  It's not hard to do manually but would lower friction for their entry to Stacks, imho.

psq:
@agraebe my setup looks different, and probably more involved, but allows for running up to date, or even patched components.
- postgres runs natively, and I manually clear/recreate database to reset
- use the docker image to run bitcoind (the only thing that really never needs updating
- run stacks-node (in mining mocknet mode) from git with suitable branch (docker images are not updated fast enough for me)
- run stacks-node-api from git from suitable branch if needed
(so that's 3 terminal windows, with full access to the logs, and easy to restart)
I set all the ports explicitly, either via environment variable or configuration rather than rely on having the startup script to pick a new one each time (try that a 100 times and let me know how if felt 😂)
I some times have more than one environment running, so again, all ports need to be set explicitly, no guessing.

I usually don't run the explorer, while a great tools when things work, it does not expose enough to diagnose problems when there are errors, so I rely on logs.

To deploy processes, I rely on my own javascript code, when working on swapr+flexr, I need to redeploy 10 contracts, in the right order (https://github.com/psq/flexr/blob/master/deploy-contracts.ts#L86-L95)

I would rather see explanations on how you can change the auth url and api url in your app code rather than rely on pseudo magic code that makes the choice for you.  In practice, it rarely does pick the ***right*** ones and explaining how to use explicit values would save people some time when it makes the wrong choice.  Plus if you have to run on a server with your custom node (auth or api), learning how to set the url explicitly is something you have to do anyway.

And I sometimes have to run with my own authenticator, so that's one more dimension to add.

Also, I would recommend people start working on their clarity contracts using the js sdk, not even mocknet, it is a lot easier to get something up and running this way (example here https://github.com/psq/flexr/blob/master/test/unit/flexr.ts, @friedger probably has simpler examples)

friedger:
@agraebe , like psq I am not using testnet explorer locally.

For mocknet, 
* I am running the docket image of stacks-blockchain-api
* I run a script to deploy my contracts and restart the docker image whenever I want to retest
* I use one or more of the four test accounts that have initial balances (with a flag in the script to switch between test accounts and my "real" testnet accounts)

@psq With the recent update of the authenticator, a local instance is not necessary anymore, see #834

psq:
> @psq With the recent update of the authenticator, a local instance is not necessary anymore, see #834

yes, I noticed, but that was necessary earlier

agraebe:
**Wow, thanks everyone for the input!** I'm updating the PR and wanted to respond to some comments

--------

> This way, you don't have to fiddle with React code. https://github.com/blockstack/ux/tree/master/packages/test-app

@hstove I looked at it and figured it would be more difficult for folks (can't resolve multiple packages, need to clone the whole ux repo, not documented as well as the other app). Sticking to an app that has a tutorial is probably best for onboarding purposes

> In my practise it is much easier to restart mocknet and redeploy the contract than finding the code for the contract addresses and replace them. In particular if you have more than one.

@friedger Good point! I added instructions for that. Instead of the script, one can restart the mocknet using docker

> The yarn lerna.. command for the authenticator should be simplified to something like yarn start-authenticator

I removed the local authenticator setup now that the network can be set

> Could the explorer be part of stacks api docker image? Or have a mocknet docker image?

Yes! Thanks to @wileyj, the [docker compose project](https://github.com/blockstack/stacks-local-dev) includes the node+API+explorer

> One nitpick ... align on a single node package manager. Instructions are mixed between npm and yarn.

Thanks @dantrevino, I fixed that!

> Setting up my own authenticator means I have to track a completely different set of different user accounts in a different set of browsers

This is not required anymore, so we can avoid that complexity

> Instructions for doing an npm init clarity-starter within an existing project would likely be easier for devs who want to bring Clarity to their existing applications.

That's true! It is possible to just run the command inside a project folder, too. However, in my instructions I can basically say that one should complete the counter tutorial before doing this one. So, I can skip over the contract creation and focus on the goal of this tutorial: setup for development.

> I usually don't run the explorer, while a great tools when things work, it does not expose enough to diagnose problems when there are errors, so I rely on logs.

@psq and @friedger Yeah, I recognized that and removed the Explorer step/requirement. It is served through the docker compose, but I don't think I explicitly need to tell devs to set it up... also, thanks for sharing your local setup details! I believe it's probably too much for the regular smart contract dev, but still good for my understanding!

> I would rather see explanations on how you can change the auth url and api url in your app code rather than rely on pseudo magic code that makes the choice for you

I updated the instructions to cover that

> Also, I would recommend people start working on their clarity contracts using the js sdk, not even mocknet, it is a lot easier to get something up and running this way

Agreed! This tutorial requires the counter tutorial to be completed, which in fact uses the clarity project generator that comes with the js sdk :) So devs can work in TDD fashion before ever deploying to mocknet. I added that to the instructions too

> I use one or more of the four test accounts that have initial balances (with a flag in the script to switch between test accounts and my "real" testnet accounts)

@friedger thanks for sharing that! The new docker compose uses test accounts and I added instructions for deployment using one of them as well (with a note so folks know about it). The script seems nice! I'm not sure it's the right place to mention it but let me think about it

agraebe:
Current blockers:

- [ ] merging https://github.com/blockstack/stacks-local-dev/pull/2

src/pages/smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,206 @@
...
+
+The initial start will take a few minutes. As soon as it is complete, you should be able to open up `[http://localhost:3999/](http://localhost:3999/)` and see:
+
+```js
+{"status":"ready"}
+```
+
+~> The Stacks Node is running on port `20443`, the Stacks Blockchain API on `3999` and the Stacks Explorer is on port `3000`.
+
+## Step 3: Deploy your contract
```
CharlieC3, line 55: @wileyj Whoops! Where's step 2?

agraebe: thanks, fixed!

aulneau (approved):

src/pages/smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,332 @@
...
+The Stacks Explorer makes it easy way to review transactions. Let's spin up a local version:
+
+```bash
+git clone https://github.com/blockstack/explorer.git
+cd explorer
+yarn
+yarn dev
+```
+
+You should now be able to open up `http://localhost:3000/` and see the homepage. **Make sure to select `Localhost` from the network list on the top right side of the screen!**
```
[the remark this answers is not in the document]
agraebe: that;d be great! did we hash out the CORS concerns that I believe were brought up last time we chatted?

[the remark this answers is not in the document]
aulneau: This will not be possible without a good amount of work. I believe the best answer is to have people run the explorer locally if they need to.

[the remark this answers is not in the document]
agraebe: makes sense, thanks!

agraebe:
@yknl this can't be merged until https://github.com/blockstack/stacks.js/issues/878 is resolved. could you take a look please?

src/pages/write-smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,206 @@
...
+
+When you develop your contract in a test-driven fashion, you will aim to complete the contract before deploying it to the testnet. Only once the contract is ready to be tested inside your app (or together with other contracts), you will need to deploy it. The fastest way to do so is by using the CLI:
+
+```shell
+# install the new CLI if not done yet
+npm install --global @stacks/cli
+
+# deploy the contract:
+# deploy_contract <SOURCE_FILE> <CONTRACT_NAME> <FEE> <NONCE> <PAYMENT_KEY> -t -T <NETWORK_ADDRESS>
+stx deploy_contract ./counter.clar counter 2000 0 cb3df38053d132895220b9ce471f6b676db5b9bf0b4adefb55f2118ece2478df01 -t -T http://localhost:20443
```
agraebe, line 67: @CharlieC3 @wileyj I tried the explorer sandbox and the CLI to deploy to the mocknet but I never get any transaction 🤔 is this image maybe outdated/broken? are you able to broadcast a tx to a mocknet?

src/pages/write-smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,201 @@
...
+
+The easiest way to spin up the Node, Blockchain API, and the Explorer locally is with Docker:
+
+```bash
+git clone https://github.com/blockstack/stacks-local-dev
+cd stacks-local-dev/
+docker-compose up -d
+```
+
+-> Make sure your Docker service is up and running
```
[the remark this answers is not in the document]
aulneau: I'd also add a bit more content around docker and docker-compose

src/pages/write-smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,201 @@
...
+
+The initial start will take a few minutes. As soon as it is complete, you should be able to open up [http://localhost:3999/](http://localhost:3999/) and see:
+
+```js
+{"status":"ready"}
+```
+
+~> The Stacks Node is running on port `20443`, the Stacks Blockchain API on `3999` and the Stacks Explorer is on port `3000`.
```
aulneau, line 53: When I pulled the latest repo, the explorer is disabled by default. we should note that.

src/pages/write-smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,201 @@
...
+- [Stacks Explorer](https://github.com/blockstack/explorer)
+
+## Step 1: Spin up mocknet
+
+The easiest way to spin up the Node, Blockchain API, and the Explorer locally is with Docker:
+
+```bash
+git clone https://github.com/blockstack/stacks-local-dev
+cd stacks-local-dev/
+docker-compose up -d
```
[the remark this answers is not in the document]
wileyj:
```
10:36:21 ~$ docker-compose up --help
...

Options:
    -d, --detach               Detached mode: Run containers in the background,
                               print new container names. Incompatible with
                               --abort-on-container-exit.
```

src/pages/write-smart-contracts/setting-up-local-environment.md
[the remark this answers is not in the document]
wileyj: what would you suggest?

src/pages/write-smart-contracts/setting-up-local-environment.md
```diff
@@ -0,0 +1,201 @@
...
+
+When you develop your contract in a test-driven fashion, you will aim to complete the contract before deploying it to the testnet. Only once the contract is ready to be tested inside your app (or together with other contracts), you will need to deploy it. The fastest way to do so is by using the CLI:
+
+```shell
+# install the new CLI if not done yet
+npm install --global @stacks/cli
+
+# deploy the contract:
+# deploy_contract <SOURCE_FILE> <CONTRACT_NAME> <FEE> <NONCE> <PAYMENT_KEY> -t -I <NETWORK_ADDRESS>
+stx deploy_contract ./counter.clar counter 2000 0 cb3df38053d132895220b9ce471f6b676db5b9bf0b4adefb55f2118ece2478df01 -t -I http://localhost:20443
```
alexkeating, line 67: Should this be -I or -T? I kept getting a NotEnoughFunds error until I found [this](https://github.com/blockstack/stacks.js/issues/840) issue and switched to -T.

agraebe: adding @yknl for confirmation but in this case, I think it should be `-I`, right? There are 3 _kind of relevant_ options (`T`, `H`, and `I`) and I'm not certain about this one

alexkeating: Okay, yea, found [this](https://github.com/blockstack/stacks.js/issues/878) and I think all my transactions are going to the testnet with -T.

agraebe:
this is quite outdated by now and there seems to be no strong demand for this type of tutorial right now. closing for now.

PR #830, opened 2020-10-19
docs: minor fixes to transaction signing sample code and content

agraebe:


src/pages/smart-contracts/signing-transactions.md
```diff
@@ -34,7 +34,7 @@ import { openContractCall } from '@blockstack/connect';
 
 // Here's an example of options:
 const myStatus = 'hey there';
-const options = {
+const opts = {
```
aulneau, line 37: I prefer the verbose `options` vs `opts`. May I ask why you changed it? (if it's because that's the interface name in connect, perhaps we should change that to `options` in place of `opts`)

src/pages/smart-contracts/signing-transactions.md
```diff
@@ -58,8 +58,7 @@ const options = {
 await openContractCall(opts);
 ```
 
-When calling `openContractCall`, you need to specify a few details. Here is the exact interface that describes what
-options you have:
+When calling `openContractCall`, you need to specify a few details. Here is the exact interface that describes what options you have:
```
aulneau, line 61:
```suggestion
There are some required parameters needed when calling `openContractCall`. Here is the exact interface which describes what options you have:
```

src/pages/smart-contracts/signing-transactions.md
agraebe: I changed it because the sample code doesn't work.. further down in the code, `opts` is used as an input for the method call

hstove: Yes, I would also prefer `options`, and then fixing the use of `opts` to be `options` lower in the doc.

hstove (changes requested):
tiny change, but thanks for this!

src/pages/smart-contracts/signing-transactions.md
agraebe: cool, done

PR #838, opened 2020-10-27
Local mocknet with docker

wileyj:
## Description
1. Super easy mocknet using docker-compose  and https://github.com/blockstack/stacks-local-dev .
2. Syntax updates for existing docker commands (removed comments from commands causing them to fail copypasta)
3. update image used for docker to use `latest` rather than a tagged version (this version is updated when a push to master occurs) - this will prevent lots of updates being required when images are updated. 

Docs being added are a close copy of  https://github.com/blockstack/stacks-local-dev/blob/master/README.md
https://github.com/blockstackpbc/devops/issues/475

### DIFF
--- a/src/common/navigation.yaml
+++ b/src/common/navigation.yaml
@@ -54,6 +54,7 @@ sections:
           - path: /transactions
           - path: /network
           - path: /stacking
+          - path: /local-development
         sections:
           - title: Tutorials
             pages:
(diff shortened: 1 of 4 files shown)


agraebe (commented):
This is great! Let me update some things in the branch directly. Most notably, I think this should be a guide and not a tutorial (and likely move to a different place in the nav)

wileyj:
> This is great! Let me update some things in the branch directly. Most notably, I think this should be a guide and not a tutorial (and likely move to a different place in the nav)

sounds good, thanks!

agraebe (approved):

Issue #837, opened 2020-10-27
PoX algorithm overview


agraebe:
A summary of the PoX algorithm that would help developers/miners/integrators understand the entire process

joshmatix:
A short summary of the Proof of Transfer whitepaper, with links to the actual White Paper and a link to the White Paper reading.

Actual figures (as in current Source Code) related to the threshold for Stackers, Initial & Sunset phases, number of blocks per Reward Cycle, maximum BTC addresses per Reward Set etc..

A graphic illustrating the interaction between the various actors and components.

agraebe:
This would be fantastic @joshmatix! Can you create a PR? Happy to help in any way possible. LMK

joshmatix:
I created a PR. Is it ok?

agraebe:
thanks @joshmatix - I'll take a look in a bit. I noticed the illustration wasn't included. maybe @jasperjansz could help as part of the upcoming docs week :)

joshmatix:
Alex,

Being new to the github process, I wanted to make sure I was doing it
right, before adding more stuff.

Tom ..

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail&utm_term=icon>
Virus-free.
www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail&utm_term=link>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

joshmatix:
Added the graphic.

PR #839, opened 2020-10-29
PoX algorithm overview

joshmatix:
## Description
Created a skeleton file for the new "PoX algorithm overview" document
For details refer to issue  #837

Not sure if PR is created at this stage or after completion.


joshmatix:
This was not done using the new docs.
Will be redone using the documented way.

PR #851, opened 2020-11-03
BNS Cleanup

kantai:
## Description

This PR cleans up the BNS content, making it applicable to BNS when Stacks 2.0 launches. It's mostly content removal, as much of the prior content is very Stacks 1.0 specific. This isn't the end of the story for BNS for Stacks 2.0 -- we should expect documentation for the BNS contract similar to the PoX contract, but the BNS contract is still in the works.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other


kyranjamie (approved):

src/pages/technology/naming-system.md
```diff
@@ -0,0 +1,405 @@
+---
+title: Overview
+description: 'Blockstack naming system (BNS)'
```
agraebe, line 3: @GinaAbrams mentioned `BNS` should now be `Blockchain naming system`

kantai: My understanding was that this name was kind of in flux, and that we should default to leaving the name as is until there's more finality.

GinaAbrams:
Thanks @kantai, think we want to get more input on this. 

cc @muneeb-ali @diwakergupta @yknl (feel free to tag more, these are just folks I remember bringing up the topic originally) 

**Naming Options:** 
SNS: Stacks Naming System 
BNS: Blockchain Naming System 

**Context from Muneeb earlier:** 
> Focusing on just the BNS vs SNS for now, having slept over it I'm leaning towards keeping it BNS. Something like "BNS built on Stacks and Bitcoin" is the angle that makes sense. Going for a more neutral "B" that stands for blockchain might be better than "S" for Stacks. Happy to discuss more.

muneeb-ali:
I'm in the BNS: Blockchain Naming System camp. Would love to hear any additional thoughts.

yknl:
I'm also in support of "Blockchain Naming System"

friedger:
There are other blockchain naming systems as well, like ens, handshake,.. naming systems with names on a blockchain

muneeb-ali:
@friedger Yep there are but we've used the acronym BNS for some years and I don't think there is currently a conflict with BNS. Also given the Bitcoin connection (e.g., a .btc namespace that is coming up) the BNS acronym can work well.

diwakergupta:
👍🏽 for "BNS (Blockchain Naming System)" though I do worry a bit that leaving it as BNS will just perpetuate the "blockstack" association.

As someone doing the heavy lifting on the actual work for BNS on Stacks 2.0, care to chime in @lgalabru ?

ps: though DNS is "domain name _system_" "BNS" in the current docs actually stands for "Blockstack Naming Service" (so service, _not_ system). Similarly ENS is "[Ethereum Name Service](https://docs.ens.domains/)".

kantai:
> ps: though DNS is "domain name system" "BNS" in the current docs actually stands for "Blockstack Naming Service" (so service, not system). Similarly ENS is "Ethereum Name Service".

I think there's some confusion in our own documentation around this -- we use System in a bunch of places (possibly my fault!) and Service in a bunch of other places (mostly in docs pages getting cleaned up here from a google search).

lgalabru:
I don't have a strong opinion, I think I'm in favor of avoid disruption coming with SNS if we think that "Bitcoin" or "Blockchain" is not a stretch.

src/common/navigation.yaml
```diff
@@ -77,24 +77,7 @@ sections:
...
-        sections:
-          - title: Tutorials
-            pages:
-              - path: /choose-name
-              - path: /create-namespace
-              - path: /resolve-name
-              - path: /register-name
-              - path: /manage-names
-              - path: /subdomains-tutorial
+      - path: /naming-system # is an overview page
```
aulneau, line 80: These need to be redirected to `/naming services`, see here https://github.com/blockstack/docs/blob/master/next.config.js#L9

aulneau (changes requested):

src/common/navigation.yaml
kantai: I see - so I should add redirects to the `next.config.js` that redirect `/naming-services/*` to `/naming-system` ?

kantai: (And fix any existing redirects)

kantai: Updated in ca79fea1

src/pages/technology/naming-system.md
kantai: Updated in ca98406c

agraebe (approved):

agraebe:
@kantai are we ready to merge or is there something outstanding?

kantai:
It looks like merging is blocked until @aulneau marks his feedback as addressed.

aulneau (approved):

PR #855, opened 2020-11-03
Remove connect page

yknl:
This PR updates the Todo app tutorial page and removes direct references to the Connect package in favour of the merged `@stacks/auth` package. The standalone Connect page has also been removed.

Should the reference content in the standalone connect page be in a API reference page instead? cc @markmhx @hstove


src/pages/authentication/connect.md
```diff
@@ -1,187 +0,0 @@
...
-
-With npm:
-
-```bash
-npm install --save @stacks/connect
-```
-
-## Usage
-
-### AuthOptions
```
agraebe, line 45: I think this chapter would be nice in some reference - maybe of the auth package?

src/pages/authentication/building-todo-app.md
```diff
@@ -95,7 +94,7 @@ export const Signin = () => {
 
 This component imports the [React hook](https://reactjs.org/docs/hooks-overview.html)
 [`useConnect`](https://github.com/blockstack/ux/blob/master/packages/connect/src/react/hooks/use-connect.ts)
-from the Stacks Connect library.
+from the Stacks Auth library.
```
markmhx, line 97: Might we want to refer to authentication here as "the Stacks authentication library", avoiding the capitalization and abbreviation "Auth", which could be interpreted as a specific product brand (e.g. "Stacks Auth")?

src/pages/authentication/building-todo-app.md
```diff
@@ -145,7 +142,7 @@ export const appConfig = new AppConfig(['store_write', 'publish_data']);
 
 The `appDetails` and `userSession` objects are joined by the callback function
 [`finished`](https://github.com/blockstack/stacks-todos/blob/master/src/components/App.jsx#L31)
-in configuring Stacks Connect for authentication with the `authOptions` object:
+in configuring Stacks Auth for authentication with the `authOptions` object:
```
markmhx, line 145: Ditto here on the "Stacks authentication" instead of "Stacks Auth" question per above.

markmhx (changes requested):

src/pages/authentication/connect.md
yknl: We have the typedoc generated reference, but that's probably not the right place for this. Perhaps in addition to the todo tutorial we need reference guide pages.

src/pages/authentication/building-todo-app.md
yknl: fixed

agraebe (approved):

markmhx (approved):

PR #856, opened 2020-11-03
Remove separate profiles page

yknl:
This PR removes the separate page about Blockstack profiles and appends the content to the auth overview page.

This page explains how to construct profiles and validate them, which is only really relevant for authenticators. Most Stacks app will never need to reference this. Therefore, it being presented as the second item in the authentication section below the overview is confusing.


agraebe (approved):
lgtm

Issue #848, opened 2020-11-02
Add documentation for helm chart


CharlieC3:
Add documentation [here](https://docs.blockstack.org/stacks-blockchain/running-testnet-node) and [here](https://docs.blockstack.org/mining) on how to run a follower and/or miner node using our stacks-blockchain helm chart.

PR #858, opened 2020-11-03
docs: add helm docs

CharlieC3:
## Description
To help make the stacks blockchain as easy as possible to run in a variety of ways, this PR adds documentation on how to run the stacks-blockchain Helm chart for the community to easily deploy a stacks node to a local or remote Kubernetes cluster.

Closes issue #848 

For details refer to PR https://github.com/blockstack/stacks-blockchain/pull/2024 (**This referenced PR must be merged before this docs PR**)

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] API reference/documentation update
- [ ] Other

No

## Checklist
- [X] `npm run lint` passes

### DIFF
--- a/src/pages/mining.md
+++ b/src/pages/mining.md
@@ -221,15 +221,13 @@ Ensure you have the following prerequisites installed on your machine:
 - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
 - [helm](https://helm.sh/docs/intro/install/)
 
-### Generate keychain
+### Generate keychain and get testnet tokens
 
 Generate a keychain:
 
 ```bash
 docker run -i node:alpine npx blockstack-cli@1.1.0-beta.1 make_keychain -t
 ```
-
-### Request testnet tokens
 
 Request BTC from the faucet:
 
--- a/src/pages/mining.md
+++ b/src/pages/mining.md
@@ -214,18 +214,22 @@ docker logs -f stacks_miner
 
 In addition, you're also able to run a testnet node in a Kubernetes cluster using the [stacks-blockchain Helm chart](https://github.com/blockstack/stacks-blockchain/tree/master/helm/stacks-blockchain).
 
--> Ensure you have the following prerequisites installed on your machine:
-  * [minikube](https://minikube.sigs.k8s.io/docs/start/) (Only needed if standing up a local Kubernetes cluster)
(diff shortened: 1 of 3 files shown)


src/pages/mining.md
```diff
@@ -209,3 +209,49 @@ You can review the node logs with this command:
...
+- [minikube](https://minikube.sigs.k8s.io/docs/start/) (Only needed if standing up a local Kubernetes cluster)
+- [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+- [helm](https://helm.sh/docs/intro/install/)
+
+### Generate keychain and get testnet tokens
+
+Generate a keychain:
+
+```bash
+docker run -i node:alpine npx blockstack-cli@1.1.0-beta.1 make_keychain -t
```
agraebe, line 229: is this the rust CLI? do we _need_ it here? all other docs use our `@stacks/cli` package for keygen. I'd love to avoid the confusion for devs if not absolutely needed

src/pages/stacks-blockchain/local-development.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Loccal development
+title: Local development
```
agraebe, line 2: 🙏

agraebe (commented):
thanks @CharlieC3! left one comment irt the CLI

src/pages/mining.md
CharlieC3: The `blockstack-cli` is used in a couple other places on this page, I kept it the same for consistency. I can update it across the board to use `@stacks/cli`, but would prefer to make that change in a separate PR.

kyranjamie (approved):

src/pages/mining.md
CharlieC3: Will update the commands to use `@stacks/cli` in my next commit

agraebe (approved):

CharlieC3:
@agraebe Not sure if I made this clear in the description, but this PR should've only be merged after https://github.com/blockstack/stacks-blockchain/pull/2024 was merged into `next`, and then into `master`.
No need to revert, since the instructions will still work, but links to the chart's github page will be broken until the referenced PR is merged into `master`.

agraebe:
ah, that was my bad! I think it's ok to keep it merged now but will pay more attention next time. I believe we could leave PRs like this as draft PRs until everything elsewhere is resolved

PR #859, opened 2020-11-04
feat: add mobile search

kyranjamie:
Refactored search state to be in global state, added icon to trigger open.

docs-mobile-search3


src/components/search-button.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const [isOpen, setIsOpen] = useState(false);
+  const onClose = () => setIsOpen(false);
+  const triggerAlgolia = () => setIsOpen(true);
+  const [colorMode] = useColorMode();
+  return (
+    <>
+      <Portal>
+        <Fade in={isOpen}>
+          {styles => (
+            <Box position="absolute" zIndex={9999} style={styles}>
```
aulneau, line 54:
```suggestion
            <Box position="fixed" zIndex={9999} style={styles}>
```

aulneau, line 54:
```suggestion
            <Box position="absolute" zIndex={9999} style={styles} css={(theme: Theme) => css({
            '.DocSearch.DocSearch-Container':{
              position:'fixed',
              }
            })(theme)}>
```

aulneau: sry for formatting, but adding `position:fixed` to `.DocSearch.DocSearch-Container` should fix it.

agraebe (approved):
just tried it out. looks & works great. will let @aulneau comment on the implementation details

aulneau:
I was able to search for an item, click on it, navigate to the page and the search modal automatically re-opened. Seems like there might be a bug in the open condition.

src/components/search.tsx
```diff
@@ -94,18 +95,19 @@ export const SearchBox: React.FC<BoxProps> = React.memo(props => {
...
 
-  const onClose = React.useCallback(
-    function onClose() {
-      setIsOpen(false);
-    },
-    [setIsOpen]
-  );
+  const onClose = React.useCallback(function onClose() {
+    setState(state => ({ ...state, searchModal: 'closed' }));
+  }, []);
```
aulneau, line 106:
```suggestion
  }, [setState]);
```

kyranjamie:
rebase squashed commits into one

kyranjamie:
Apologies for letting this PR go stale. 

I've really no idea what's causing the double opening. 
- It doesn't happen locally, only when deployed
- I can't find any triggers that would cause the boolean flag to switch back to true

agraebe:
@aulneau could you take a look please?

aulneau (approved):

PR #871, opened 2020-11-05
feat: add stacks-node documentation

### LINKED ISSUE #869: Add docs for stacks-node subcommands and config options
Add documentation on different subcommands and configuration options available for the `stacks-node` binary.

CharlieC3:
## Description

Adds documentation on different subcommands and configuration options available for the `stacks-node` binary.

@kantai & @lgalabru: I'm looking for feedback particularly on the explanations I gave for each subcommand and configuration option, examples used, formatting, etc. Please let me know if there are any subcommands and/or config options I should leave out because they'll be removed soon, or any that I may have missed.

Closes #869 

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [X] API reference/documentation update
- [ ] Other

No

## Checklist
- [X] `npm run lint` passes

### DIFF
--- a/src/common/navigation.yaml
+++ b/src/common/navigation.yaml
@@ -54,6 +54,7 @@ sections:
           - path: /network
           - path: /stacking
           - path: /local-development
+          - path: /configurations
         sections:
           - title: Tutorials
             pages:
--- a/src/common/navigation.yaml
+++ b/src/common/navigation.yaml
@@ -54,7 +54,6 @@ sections:
           - path: /network
           - path: /stacking
           - path: /local-development
-          - path: /configurations
         sections:
           - title: Tutorials
             pages:
@@ -116,6 +115,7 @@ sections:
           - external:
               href: 'https://blockstack.github.io/stacks.js/'
               title: Stacks.js
+          - path: /node-configuration
           - path: /faqs
           - path: /glossary
           - path: /deploy-tips
--- a/src/common/navigation.yaml
+++ b/src/common/navigation.yaml
@@ -112,10 +112,10 @@ sections:
           - path: /language-clarity
           - path: /stacking-contract
           - path: /stacks-blockchain-api
+          - path: /stacks-node-configuration
           - external:
               href: 'https://blockstack.github.io/stacks.js/'
               title: Stacks.js
-          - path: /node-configuration
           - path: /faqs
           - path: /glossary
           - path: /deploy-tips
--- a/src/pages/references/node-configuration.md
+++ b/src/pages/references/node-configuration.md
@@ -227,6 +227,18 @@ Example:
 miner = true
 ```
 
+#### mine_microblocks (optional)
+
+Determines whether the stacks-node will mine microblocks. Will only take effect if `miner` is set to `true`.
+
+[See this page for information on how to run a miner.](https://docs.blockstack.org/mining)
+
+Example:
+
+```toml
+mine_microblocks = true
+```
+
 #### prometheus_bind (optional)
 
 Address and port stacks-node should open for Prometheus metrics collection.
--- a/src/pages/references/node-configuration.md
+++ b/src/pages/references/node-configuration.md
@@ -97,7 +97,7 @@ miner = true
 
 #### working_dir (optional)
 
-Absolute path to the directory which the stacks-node will use for storing various data
+Absolute path to the directory which the stacks-node will use for storing various data.
 
 Example: `"/root/stacks-node"`
 
@@ -215,19 +215,19 @@ walk_interval = 30
 
 #### public_ip_address
 
-The advertised public IP of this stacks-node
+The advertised public IP of this stacks-node.
 
 Example: `"1.2.3.4:20444"`
 
 #### download_interval
 
-Time (in seconds) between attempts to download blocks
+Time (in seconds) between attempts to download blocks.
 
 Example: `10`
 
 #### walk_interval
 
-Time (in seconds) between attempts to walk the Merkle tree
+Time (in seconds) between attempts to walk the Merkle tree.
 
 Example: `30`
 
@@ -335,4 +335,5 @@ Example: `"STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6"`
 #### amount
 
 The balance of micro-STX given to the address.
+
 Example: `10000000000000000`
--- a/src/pages/references/node-configuration.md
+++ b/src/pages/references/node-configuration.md
@@ -7,6 +7,18 @@ images:
   sm: /images/pages/testnet-sm.svg
 ---
 
+## Table of Contents
+
+- [Installation](#installation)
+- [Usage](#usage)
+- [Subcommands](#subcommands)
+- [Configuration File Options](#configuration-file-options)
+  - [Section: node](#section-node)
+  - [Section: events_observer (optional)](#section-events_observer-optional)
+  - [Section: connection_options (optional)](#section-connection_options-optional)
+  - [Section: burnchain](#section-burnchain)
+  - [Section: mstx_balance](#section-mstx_balance)
+
 ## Installation
 
 See this page for installation instructions.
@@ -69,7 +81,7 @@ To see a list of example configurations, [please see this page](https://github.c
 
 ### Section: node
 
-This section contains configuration options pertaining to the stacks-node.
+Contains various configuration options pertaining to the stacks-node.
 
 Example:
 
@@ -119,6 +131,12 @@ Public key, address, and port stacks-node should use to pull transaction data fr
 
 Example: `"048dd4f26101715853533dee005f0915375854fd5be73405f679c1917a5d4d16aaaf3c4c0d7a9c132a36b8c5fe1287f07dad8c910174d789eb24bdfb5ae26f5f27@krypton.blockstack.org:20444"`
 
+#### wait_time_for_microblocks (optional)
+
+Time to wait (in milliseconds) for microblocks.
+
+Example: `15000`
+
 #### seed (optional)
 
 The private key to use for mining. Only needed if `miner` is set to `true`.
@@ -151,6 +169,8 @@ Example: `"0.0.0.0:9153"`
 
 ### Section: events_observer (optional)
 
+Contains options for watching events emitted by a local [stacks-blockchain-api](https://github.com/blockstack/stacks-blockchain-api) service.
+
 -> This section can be repeated multiple times.
 
 Example:
@@ -179,6 +199,37 @@ Example: `255`
 Event keys for which to watch.
 
 Example: `["*"]`
+
+### Section: connection_options (optional)
+
+Specifies configuration options for others connecting to the stacks node.
+
+Example:
+
+```toml
+[connection_options]
+public_ip_address = "1.2.3.4:20444"
+download_interval = 10
+walk_interval = 30
+```
+
+#### public_ip_address
+
+The advertised public IP of this stacks-node
+
+Example: `"1.2.3.4:20444"`
+
+#### download_interval
+
+Time (in seconds) between attempts to download blocks
+
+Example: `10`
+
+#### walk_interval
+
+Time (in seconds) between attempts to walk the Merkle tree
+
+Example: `30`
 
 ### Section: burnchain
 
@@ -245,7 +296,7 @@ Example: `30000`
 
 #### commit_anchor_block_within (optional)
 
-Sets the time period for commitments. Only used when `mode` is set to `"helium"`.
+Sets the time period (in milliseconds) for commitments. Only used when `mode` is set to `"helium"`.
 
 Example: `10000`
 
--- a/src/pages/references/node-configuration.md
+++ b/src/pages/references/node-configuration.md
@@ -27,7 +27,9 @@ See this page for installation instructions.
 
 ## Usage
 
-`stacks-node sub-command [--subcommand-option <value>]`
+```bash
+stacks-node sub-command [--subcommand-option <value>]
+```
 
 ## Subcommands
 
@@ -35,19 +37,31 @@ See this page for installation instructions.
 
 Start a node based on a fast local setup emulating a burnchain. Ideal for smart contract development.
 
-Example: `stacks-node mocknet`
+Example:
+
+```bash
+stacks-node mocknet
+```
 
 ### krypton (deprecated)
 
 Start a node that will join and stream blocks from the public krypton testnet, powered by Blockstack via (Proof of Transfer).
 
-Example: `stacks-node krypton`
+Example:
+
+```bash
+stacks-node krypton
+```
 
 ### xenon
 
 Start a node that will join and stream blocks from the public xenon testnet, decentralized.
 
-Example: `stacks-node xenon`
+Example:
+
+```bash
+stacks-node xenon
+```
 
 ### start
 
@@ -57,7 +71,11 @@ Start a node with a config of your own. Can be used for joining a network, start
 
 **--config**: relative or absolute path to the TOML config file. Required.
 
-Example: `stacks-node start --config=/path/to/config.toml`
+Example:
+
+```bash
+stacks-node start --config=/path/to/config.toml
+```
 
 See [Configuration File Options](#configuration-file-options) for more information.
 
@@ -65,13 +83,21 @@ See [Configuration File Options](#configuration-file-options) for more informati
 
 Displays information about the current version and the release cycle.
 
-Example: `stacks-node version`
+Example:
+
+```bash
+stacks-node version
+```
 
 #### help
 
 Displays a help message.
 
-Example: `stacks-node help`
+Example:
+
+```bash
+stacks-node help
+```
 
 ## Configuration File Options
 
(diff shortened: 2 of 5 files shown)


src/pages/stacks-blockchain/configurations.md
```diff
@@ -0,0 +1,282 @@
...
+
+#### burn_fee_cap
+
+Max burn fee for a transaction.
+
+Example: `30000`
+
+#### commit_anchor_block_within
+
+TODO
```
CharlieC3, line 245: @lgalabru @kantai What is this field for?

kantai: It's used by `helium` nodes (i.e., local-only miners) to set the time period for commitments.

src/common/navigation.yaml
```diff
@@ -54,6 +54,7 @@ sections:
           - path: /network
           - path: /stacking
           - path: /local-development
+          - path: /configurations
```
agraebe, line 57: could we rename the file so it is more descriptive in the navigation? I'd propose something like `node-configuration`

src/pages/stacks-blockchain/configurations.md
```diff
@@ -0,0 +1,284 @@
+---
+title: Configurations
```
agraebe, line 2: `Node configuration`

src/pages/stacks-blockchain/configurations.md
```diff
@@ -0,0 +1,284 @@
...
+
+See this page for installation instructions.
+[@page-reference | inline]
+| /stacks-blockchain/running-testnet-node
+
+## Usage
+
+`stacks-node sub-command [--subcommand-option <value>]`
+
+## Command-line Subcommands and Options
```
agraebe, line 20: This reads like a reference page - could we maybe move this to the reference section instead?

src/common/navigation.yaml
agraebe: update: with this looking more like a reference, maybe it is better to name it `stacks-node` or similar (so it would be accessible as: "references > stacks node"). Would suggest a slightly broader scope, but might be good to have a space for node references in general

src/pages/stacks-blockchain/configurations.md
CharlieC3: Totally! I didn't think of that. Updated.

src/pages/stacks-blockchain/configurations.md
CharlieC3: Updated

src/common/navigation.yaml
CharlieC3: [What do you think of this?](https://3000-a973afd8-e39b-4e4c-9108-9aeb922360b9.ws-us02.gitpod.io/) (Using my fancy gitpod)

src/pages/stacks-blockchain/configurations.md
CharlieC3: Added description, thanks!

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,338 @@
...
+
+Time (in seconds) between attempts to download blocks
+
+Example: `10`
+
+#### walk_interval
+
+Time (in seconds) between attempts to walk the Merkle tree
+
+Example: `30`
```
CharlieC3, line 232: Took a swing at these descriptions. How accurate are they?

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,477 @@
...
+
+Start a node based on a fast local setup emulating a burnchain. Ideal for smart contract development.
+
+Example:
+
+```bash
+stacks-node mocknet
+```
+
+### krypton (deprecated)
```
CharlieC3, line 46: I intentionally left out the `helium` and `argon` subcommands since we're past those testnet phases, and put `deprecated` for the `krypton` subcommand since this testnet phase will be completed soon. Was this the right move?

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Start a node that will join and stream blocks from the public krypton testnet, powered by Blockstack via (Proof of Transfer).
```
friedger, line 48: ..powered by Stacks via

Why is "Proof of Transfer" in brackets?

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```bash
+stacks-node krypton
+```
+
+### xenon
+
+Start a node that will join and stream blocks from the public xenon testnet, decentralized.
```
friedger, line 58: "decentralized" is not clear to me.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```bash
+stacks-node xenon
+```
+
+### start
+
+Start a node with a config of your own. Can be used for joining a network, starting new chain, etc.
```
friedger, line 68: Can be used for joining a network, starting new chain, replacing default values for `mocknet` or `xenon`.

src/pages/references/node-configuration.md
friedger: I assume that it will be possible to configure the testnet xenon to run with a single node for testing.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+bootstrap_node = "048dd4f26101715853533dee005f0915375854fd5be73405f679c1917a5d4d16aaaf3c4c0d7a9c132a36b8c5fe1287f07dad8c910174d789eb24bdfb5ae26f5f27@krypton.blockstack.org:20444"
+```
+
+#### wait_time_for_microblocks (optional)
+
+Time to wait (in milliseconds) for microblocks.
```
friedger, line 186: Not clear to me what it means to wait for microblocks.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+seed = "replace-with-your-private-key"
+```
+
+#### local_peer_seed (optional)
+
+The private key to use for mining.
```
friedger, line 208: What is the difference to `seed`?

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+local_peer_seed = "replace-with-your-private-key"
+```
+
+#### miner (optional)
+
+Determines whether the stacks-node is running a follower (`false`) or a miner (`true`).
```
friedger, line 220: Default value is `false`

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+endpoint = "address-to-my-local.stacks-node-api.com:3700"
+```
+
+#### retry_count
+
+Number of times to retry before failing.
```
friedger, line 279: Number of times to retry sending events to the endpoint before failing

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+retry_count = 255
+```
+
+#### events_keys
+
+Event keys for which to watch.
```
friedger, line 289: The emitted node events can be restricted by account, function name and event type. Asterix ("*") can be used to emit all events.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+events_keys = ["*"]
```
friedger, line 294:
```toml
events_keys = [
    "STGT7GSMZG7EA0TS6MVSKT5JC1DCDFGZWJJZXN8A.store::print",
    "STGT7GSMZG7EA0TS6MVSKT5JC1DCDFGZWJJZXN8A.contract.ft-token",
    "STGT7GSMZG7EA0TS6MVSKT5JC1DCDFGZWJJZXN8A.contract.nft-token",
     "stx"
]
```

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+download_interval = 10
+```
+
+#### walk_interval
+
+Time (in seconds) between attempts to walk the Merkle tree.
```
friedger, line 332: to walk the neighborhood.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+```toml
+[burnchain]
+chain = "bitcoin"
+mode = "krypton"
+peer_host = "bitcoind.blockstack.org"
+rpc_port = 18443
+peer_port = 18444
+```
+
+This section contains configuration options pertaining to the blockchain the stacks-node binds to on the backend for proof-of-transfer (BTC).
```
friedger, line 353: Description first, then example.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+#### chain
+
+The blockchain stacks-node binds to on the backend for proof-of-transfer.
```
friedger, line 357: Only value supported: "bitcoin"

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+chain = "bitcoin"
+```
+
+#### mode
+
+The profile of which to run stacks-node.
```
friedger, line 367: "profile", "mode" and sub commands are used to describe the same thing, however, it is not clear to me what the thing is.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+mode = "krypton"
+```
+
+#### peer_host
+
+Host running the BTC blockchain.
```
friedger, line 377: Just "Host running the blockchain" ?

src/pages/references/node-configuration.md
friedger: What are the valid values?

src/common/navigation.yaml
```diff
@@ -115,6 +115,7 @@ sections:
           - external:
               href: 'https://blockstack.github.io/stacks.js/'
               title: Stacks.js
+          - path: /node-configuration
```
agraebe, line 118: Just looked at the preview and I think "stacks-node-config` would be best for display in the navigation. Just "node" inside the references might not be enough for someone scanning through the navigation items

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+commit_anchor_block_within = 10000
+```
+
+### Section: mstx_balance
+
+This section contains configuration options pertaining to the micro-STX balanaces for various addresses.
```
friedger, line 447: "pertaining to account balances for an address in micro-STX. This is used to define balances in the genesis block."

One section is for one address.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
+---
+title: Node Configuration
+description: Configuration parameters and options for the stacks-node binary
+icon: TestnetIcon
+images:
+  large: /images/pages/testnet.svg
+  sm: /images/pages/testnet-sm.svg
+---
+
+## Table of Contents
```
agraebe, line 10: we actually preview the contents on the right-side page nav. I'd remove the ToC from the markdown

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Example:
+
+```toml
+address = "STB44HYPYAT2BB2QE513NSP81HTMYWBJP02HPGK6"
+```
+
+#### amount
+
+The balance of micro-STX given to the address.
```
friedger, line 483: given to the address at start of the node.

src/pages/references/node-configuration.md
agraebe: I don't have the full understanding of the config options but my general suggestion would be to not assume folks know what phase we're currently in. So the heading based on the phases would likely be very confusing. Maybe we can add the deprecation node to the specific commands/options? 

To answer the other question - deprecated content (assuming it is not relevant anymore) shouldn't be added to the docs. would rather increase confusion

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+stacks-node help
+```
+
+## Configuration File Options
+
+The TOML configuration file has multiple sections under which an option may be placed.
+
+To see a list of example configurations, [please see this page](https://github.com/blockstack/stacks-blockchain/tree/master/testnet/stacks-node/conf).
+
+### Section: node
```
agraebe, line 108: I'd remove the `Section: ` - we're not doing that anywhere else

friedger (changes requested):
Left some comments. How to provide default values and valid values? For a few properties a value is given, for others not.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+- [Usage](#usage)
+- [Subcommands](#subcommands)
+- [Configuration File Options](#configuration-file-options)
+  - [Section: node](#section-node)
+  - [Section: events_observer (optional)](#section-events_observer-optional)
+  - [Section: connection_options (optional)](#section-connection_options-optional)
+  - [Section: burnchain](#section-burnchain)
+  - [Section: mstx_balance](#section-mstx_balance)
+
+## Installation
```
agraebe, line 22: Now that this page is a reference, we can probably remove this installation piece. For best results, it would likely be nice to update the tutorial referenced here though. Something at the end of the tutorial that would say "for a full reference of Stacks Node commands and config options, check out ..."

src/pages/references/node-configuration.md
CharlieC3: I labeled these parts as "sections" because in the config file's format, TOML, each string between brackets `[]` is called a table or sometimes a section. It's specific to the TOML format, and I feel removing it would make each section appear as a valid config field when in fact it's a table/section which may contain a particular set of fields.

I'm open to ideas on how to word/structure this differently, but I feel it needs to differentiate itself from the config fields, and frustratingly the `####` heading appears larger than the `###` heading.

src/pages/references/node-configuration.md
CharlieC3: heh, I meant to link it to the doc on PoX. Good catch!

src/common/navigation.yaml
agraebe: 👍

src/pages/references/node-configuration.md
agraebe: I see, in that case, you are right. Let's keep it :)

src/pages/references/node-configuration.md
CharlieC3: Is there any way to customize the preview on the right side? It only lists the top-level `##` headings, and if I can get it to resemble my ToC that would be perfect.

src/pages/references/node-configuration.md
CharlieC3: I might remove it then, it's not clear to me either. It's just the description I pulled from the `stacks-node` help message.

src/pages/references/node-configuration.md
CharlieC3: Even though we're now past some phases, these are actual subcommands to the `stacks-node` binary. I think we just haven't had the time to remove them from code yet. So while I could remove them from the docs, users will still see them appear in the `stacks-node`'s help message until it's removed in code.

I agree on the deprecation point, just didn't want this to come across as "hey the PBC is hiding info!", and instead just offer the docs based on the current status of the `stacks-node` binary.

src/pages/references/node-configuration.md
```diff
@@ -0,0 +1,489 @@
...
+
+Sets the time period (in milliseconds) for commitments. Only used when `mode` is set to `"helium"`.
+
+Example:
+
+```toml
+commit_anchor_block_within = 10000
+```
+
+### Section: mstx_balance
```
kantai, line 445: One thing I would add here -- these micro-STX balances define the "genesis block allocations" for the blockchain / testnet. If a user _changes_ these values, their node may be in conflict with other nodes on the network and find themselves unable to sync with other nodes (because they are in fact mining a fork).

kantai (approved):
This looks great to me, thanks @CharlieC3, just had one comment.

src/pages/references/node-configuration.md
CharlieC3: @kantai or @lgalabru any insight on this one?

src/pages/references/node-configuration.md
agraebe: that makes sense! For usability reasons, I'd focus on documenting what we'd like folks to use as opposed to everything there is. Maybe we could add a note to the references explicitly saying that the binary includes other commands that are deprecated but not yet removed:

```
-> Note that the `stacks-node` binary could include deprecated commands that are not documented on this page. Deprecated commands could be accessible until they are fully removed from the sources
```

src/pages/references/node-configuration.md
agraebe: unfortunately, we don't have an option to set this per page. the two-level heading is critical for longer pages and seemed a good balance overall across the docs. I wouldn't worry about this too much. I'd assume that devs going to the sources would control+f anyways

src/pages/references/node-configuration.md
CharlieC3: Valid values can be found [here](https://github.com/blockstack/stacks-blockchain/blob/master/testnet/stacks-node/src/config.rs#L431), I'll include it in the description.

I said "profile" just to avoid re-using the word "mode" in the definition. I'm assuming this is a kind of feature toggle, and the subcommands effectively select the `mode` behind the scenes in the config they load in code.

agraebe (approved):
lgtm

src/pages/references/node-configuration.md
CharlieC3: Not sure, @kantai or @lgalabru ?

kantai: This is the seed used for signing P2P messages in the networking stack, it differentiates network peers (and is used even by non-mining nodes).

src/pages/references/node-configuration.md
kantai: This is the amount of time that a node will wait before trying to mine a block, after catching up to the anchored chain tip. This gives the current leader time to broadcast microblocks that will get included in that mined block.

Issue #816, opened 2020-10-09
Guide for Clarity values


agraebe:
The Clarity Type system is mostly documented in the references. We should better document types in the smart contract category. This guide should include:

- [ ] Type description for all types (including string)
- [ ] Type declaration / instantiation
- [ ] Type samples
- [ ] Encoding/Decoding details
- [ ] Library support
- [ ] Helper methods (cvToString, )

friedger:
This might be helpful: https://github.com/blockstack/stacks.js/tree/master/packages/transactions#constructing-clarity-values

PR #877, opened 2020-11-06
Clarity Values Guide

reedrosenbluth:
A guide for dealing with Clarity values in JS.

#816

### DIFF
--- a/src/pages/smart-contracts/clarity-values-guide.md
+++ b/src/pages/smart-contracts/clarity-values-guide.md
@@ -1,6 +1,6 @@
 ---
 title: Clarity Values Guide
-description: Learn how to deal with Clarity Values in Javascript.
+description: Learn how to deal with Clarity Values in JavaScript.
 tags:
   - tutorial
 ---
--- a/src/pages/smart-contracts/clarity-values-guide.md
+++ b/src/pages/smart-contracts/clarity-values-guide.md
@@ -8,7 +8,7 @@ tags:
 ## Introduction
 The Clarity language makes use of a strong static [type system](https://docs.blockstack.org/references/language-clarity#clarity-type-system). This simply means that every function defined in Clarity expects arguments of specific types, and that a failure to provide properly typed arguments will result in your code failing to compile, or your `contract-call` transaction failing prior to execution.
 
-In order to build web applications that interact with Clarity contracts, we must include type information as a part of our function calls. The [@stacks/transactions](https://github.com/blockstack/stacks.js/tree/master/packages/transactions) library makes this easy, as we will demonstrate below.
+In order to build web applications that interact with Clarity contracts, you will need to learn how to construct and use `ClarityValues`. The [@stacks/transactions](https://github.com/blockstack/stacks.js/tree/master/packages/transactions) library makes this easy, as we will demonstrate below.
 
 ## Clarity Types
 Values in Clarity can have the following types:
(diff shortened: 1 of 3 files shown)


src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
+---
+title: Clarity Values Guide
+description: Learn how to deal with Clarity Values in Typescript.
+tags:
+  - tutorial
```
reedrosenbluth, line 5: Should there be a new sidebar category for "Guides"?

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
+---
+title: Clarity Values Guide
+description: Learn how to deal with Clarity Values in Javascript.
```
kyranjamie, line 3:
```suggestion
description: Learn how to deal with Clarity Values in JavaScript.
```

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
+---
```
agraebe, line 1: this page isn't in the navigation yet. can you make a new entry here: https://github.com/blockstack/docs/blob/master/src/common/navigation.yaml#L8

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+title: Clarity Values Guide
```
agraebe, line 2: I'd just say "Clarity Values" - we don't specifically label guides in the title

src/pages/smart-contracts/clarity-values-guide.md
agraebe: yeah, that's a good question! We currently seem to have 3 types of content: guides, tutorials, and references. Guides are assumed the "default" type and are not labeled right now. tutorials are labeled (tag + tutorial category in the nav). references are in their own category. We probably should revisit the behavior at some point but I'd not worry about it as part of the docs week

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+- `(buff max-len)` - byte buffer or maximum length max-len.
+- `(string-ascii max-len)` - ASCII string of maximum length max-len
+- `(string-utf8 max-len)` - UTF-8 string of maximum length max-len
+- `principal` - object representing a principal (whether a contract principal or standard principal).
+- `bool` - boolean value (true or false)
+- `int` - signed 128-bit integer
+- `uint` - unsigned 128-bit integer
+
+## Constructing Clarity Values
+Clarity values can be constructed with functions provided by the **@stacks/transactions** library. These function simply output javascript objects that contain a value and a numerical representation of the Clarity type information (or in the case of Booleans and Optional None, just the type information). These Clarity value objects can then be easily serialized into `contract-call` transactions for interacting with Clarity contracts.
```
agraebe, line 29: `These functions [...]`

agraebe: "contract-call transactions" > "contract call transactions" (assuming the reader does not yet know that is the type (might be a bit confusing). I'd also reference something like this chapter: https://docs.blockstack.org/stacks-blockchain/transactions#construction

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+const address = 'SP2JXKMSH007NPYAQHKJPQMAQYAD90NQGTVJVQ02B';
+const contractName = 'contract-name';
+
+const spCV = standardPrincipalCV(address);
+const cpCV = contractPrincipalCV(address, contractName);
+```
+
+### Response Values
+```javascript
+const errCV = responseErrorCV(trueCV());
```
agraebe, line 73: Is it correct to assume that folks can't know the response type ahead of the call completion? In that case, we should explain as well how the response can be identified (ie is this an error or OK response)?

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Buffers
+```javascript
+const buffer = Buffer.from('foo');
+const bufCV = bufferCV(buffer);
+```
+
+### Integers
+```javascript
+const i = intCV(-10);
```
agraebe, line 52: does it make sense to talk about BN briefly? How should these be used together (or not)?

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Lists
+```javascript
+const l = listCV([trueCV(), falseCV()])
+```
+
+## Using Clarity Values
+Now that you know how to construct Clarity values, you can use them to build `contract-call` transactions that call smart contract functions.
+
+Here's an example of how this is done:
```
agraebe, line 94: To avoid duplicated content, maybe we should just link to this chapter: https://docs.blockstack.org/stacks-blockchain/transactions#construction ?

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+
+const transaction = await makeContractCall(txOptions);
+
+broadcastTransaction(transaction, network);
+```
+
+## Utilizing Clarity Values from Transaction Responses
+If you successfully submit a `contract-call` transaction that calls a Clarity function, you may receive a resulting Clarity value in response.
+
+If using the transaction library's `callReadOnlyFunction()` method, the result will be a `ClarityValue`. As mentioned above, `ClarityValues` are simply javascript objects containing a value and its associated Clarity type information. These object types are defined [here](https://github.com/blockstack/stacks.js/tree/1f2b5fd8bdf1c2b5866e8171163594d7708a8c7a/packages/transactions/src/clarity/types).
```
agraebe, line 125: let's link to a reference for the `callReadOnlyFunction`

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Deserializing Clarity Values from Hex
```
agraebe, line 127: can we also talk about `deserializeCV` and `serializeCV`?

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+If you receive a response from a transaction in the form of a Hex string, you can deserialize it into a Clarity value like so:
+
+```javascript
+import { hexToCV } from '@stacks/transactions';
```
agraebe, line 131: should we also explain that the types have toString methods that could be used to convert Clarity to hex?

```
uintCV(numberOfCycles)).toString('hex')}
```

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+const bufCV = bufferCV(buffer);
+```
+
+### Integers
+```javascript
+const i = intCV(-10);
+const u = uintCV(10);
+```
+
+### Strings
```
aulneau, line 56: it'd be nice for all these CV's to have an example of the output:
```javascript
const ascii = stringAsciiCV('hello world');
// output
const utf8 = stringUtf8CV('hello 🌾');
// output
```

yknl (commented):
I think it would be really helpful to document the reverse of constructing Clarity values. For example, if I receive a responseCV from a contract function call. How do I extract the value and how do I work with them in JavaScript?

src/pages/smart-contracts/clarity-values-guide.md
reedrosenbluth: What exactly do you mean by "output"? Do you mean the JS objects that are created by these functions? They aren't very pretty to look at unless they are serialized into strings using `cvToString()` 😅

src/pages/smart-contracts/clarity-values-guide.md
reedrosenbluth: The functions are listed directly below

src/pages/smart-contracts/clarity-values-guide.md
reedrosenbluth: `Response` is a single type, with two possible values, `Ok(val)` and `Err(val)`.

Definitely safe and correct to assume that it won't be possible to anticipate if a successful response or an error response will be returned by a function that returns a `Response`.

I think I'll add a section below on how to detect what type a Clarity Value is

src/pages/smart-contracts/clarity-values-guide.md
reedrosenbluth: `intCVs` are represented internally as BNs, but the user doesn't necessarily need to know about that since the `intCV()` function simply takes an int as an argument

src/pages/smart-contracts/clarity-values-guide.md
reedrosenbluth: They actually don't have `toString()` methods

agraebe: oh, is it inherited from the base class? I think we should mention the toString usage for hex _somewhere_ at least

reedrosenbluth: There is no base class. All `ClarityValues` are simply plain javascript objects.

The recommended way to convert them to strings is using the `cvToString()` method, which deals with obtaining the the underlying data and converting it to a `hex` string under the hood.

src/pages/smart-contracts/clarity-values-guide.md
reedrosenbluth: Are these functions really important enough to be mentioned in this guide? When would a dev need to use them?

My understanding is that developers will mainly be serializing/deserializing entire transactions, not individual Clarity Values. Is this incorrect?

agraebe: I think contract calls from JS would need to use these methods, see the [stacking tutorial](https://docs.blockstack.org/stacks-blockchain/integrate-stacking#step-4-verify-stacking-eligibility)

agraebe (approved):

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Introduction
+The Clarity language makes use of a strong static [type system](https://docs.blockstack.org/references/language-clarity#clarity-type-system). This simply means that every function defined in Clarity expects arguments of specific types, and that a failure to provide arguments of the specified types will result in your code failing to compile, or your `contract-call` transaction failing prior to execution.
+
+In order to build web applications that interact with Clarity contracts, we must include type information as a part of our function calls. The [@stacks/transactions](https://github.com/blockstack/stacks.js/tree/master/packages/transactions) library makes this easy, as we will demonstrate below.
+
+## Clarity Types
+The Clarity type system includes the following types:
+
+- `(tuple (key-name-0 key-type-0) (key-name-1 key-type-1) ...)` - a typed tuple with named fields.
```
njordhov, line 16: This type corresponds to the ECMAScript [Record](https://github.com/tc39/proposal-record-tuple), an immutable mapping from keys to values. The `tuple` type in Clarity is a [misnomer](https://github.com/blockstack/stacks-blockchain/issues/1520), likely to confuse JS devs as TC39 will also introduce a `tuple` datatype in ECMAScript that is notably different than the `tuple` in Clarity.

yknl (approved):
LGTM, the additional explanation on utilizing Clarity objects in javascript is helpful.

aulneau (approved):

src/pages/smart-contracts/clarity-values-guide.md
agraebe: The filename of this file needs to match the entry in the yaml file. Should update it to `clarity-values`

src/pages/smart-contracts/clarity-values-guide.md
```diff
@@ -0,0 +1,160 @@
+---
+title: Clarity Values Guide
+description: Learn how to deal with Clarity Values in JavaScript.
+tags:
+  - tutorial
```
agraebe, line 5: because this is not in the tutorial section of the smart contract chapter, we should remove the tutorial tag here

PR #882, opened 2020-11-06
Add redirects for wallet pages and remove from navigation

markmhx:
Resolves https://github.com/blockstack/docs/issues/879


next.config.js
```diff
@@ -340,7 +384,6 @@ async function redirects() {
       destination: '/storage-hubs/gaia-admin',
       permanent: true,
     },
-    { source: '/org/secureref.html', destination: '/stacks-wallet/security', permanent: true },
```
agraebe, line 343: I'm assuming it's not a good practice to remove existing URLs unless we're certain they aren't referenced anywhere. I'd suggest that we point them to the wallet homepage

agraebe (commented):
looking great. I left a note irt redirect removal (form the old docs URL structure)

next.config.js
markmhx: Git is tracking this as a deletion but it's actually moved and updated further up on [line 298](https://github.com/blockstack/docs/pull/882/files#diff-882b2c04b01b2e8b2cdcf1817c30ea503a8005f1c0d54cfff37053b6801fea85R298).

agraebe (approved):
awesome, lgtm!

agraebe:
@markmhx could you review and resolve the conflicts? Ping me if I should help

PR #905, opened 2020-11-11
feat: proposal for an interactive tour of Clarity

lgalabru:
## Description

This PR is a proposal for introducing an interactive tour of Clarity, based on Jupyter notebooks. 

Right now, the notebooks are hosted on a free service, so the initialization is pretty slow, but we can probably address this problem by throwing some money at it.

Concerning the look and feel, Jupyter notebooks are by default rudimentary and not very appealing, but @aulneau and I talked a few months ago, and I think there are ways to tweak this.  

Concerning the content, it's still very much draft / work in progress. If we think that this approach make sense, then we can iterate on this and ship it.

### DIFF
--- a/src/pages/smart-contracts/interactive-tour.md
+++ b/src/pages/smart-contracts/interactive-tour.md
@@ -1,0 +1,83 @@
+---
+title: Interactive Tour
+description: Learn the basics of Clarity.
+duration: 18 minutes
+experience: beginners
+tags:
+  - tutorial
+images:
+  large: /images/pages/hello-world-app.svg
+  sm: /images/pages/hello-world-sm.svg
+---
+
+## Introduction
+
+In this section, you learn about Clarity language reference. By the end of the tutorial, you will be familiar with:
+
+- Functional programming basics,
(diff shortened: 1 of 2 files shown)


whoabuddy:
Would this be something we can pair with https://clarity.tools/ rather than Jupyter?

aulneau:
@whoabuddy I am currently working on an inline clarity repl implementation :)

agraebe:
what should be the next step here?

lgalabru:
@agraebe I'm not sure. If you're in the process of "marie kondo"-ing the PRs in that repo, then we can open an issue, close that PR, and re-evaluate once we have bandwidth for the notebooks?

lgalabru:
> Would this be something we can pair with https://clarity.tools/ rather than Jupyter?

Probably. Clarity.tools not being source, it's hard to know what does the underlying format looks like.

agraebe:
> in the process of "marie kondo"-ing the PRs

haha, I've never heard that name but I heard of the book. yeah, let's do what you suggested!

PR #916, opened 2020-11-17
feat: update btc faucet location

hstove:
This is a small change to point miners to a different testnet BTC faucet. This is required for Xenon, because we'll no longer be using our own testnet.

Don't merge until Xenon is ready!

### DIFF
--- a/src/pages/mining.md
+++ b/src/pages/mining.md
@@ -47,14 +47,9 @@ After this runs, you'll probably see some installation logs, and at the end you 
 }
 ```
 
-We need to get some testnet BTC to that address. Grab the `btcAddress` field, and call the BTC faucet:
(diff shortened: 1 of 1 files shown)


src/pages/mining.md
```diff
@@ -47,14 +47,9 @@ After this runs, you'll probably see some installation logs, and at the end you
 }
 ```
 
-We need to get some testnet BTC to that address. Grab the `btcAddress` field, and call the BTC faucet:
-
-```bash
-# replace <btc_address> with `btcAddress` property from your keychain
-curl -XPOST "https://stacks-node-api.blockstack.org/extended/v1/faucets/btc?address=<btc_address>" | json_pp
```
agraebe, line 54: I looked at the file and the `/btc` endpoint is mentioned a couple of more times - should these be updated as well?

agraebe: @hstove ping

agraebe (approved):
lgtm

PR #925, opened 2020-11-22
Replace blockstack-cli references

alexkeating:
Use the following template to create your pull request

## Description

I replaced all the blockstack-cli references in the stacks cli reference page. This burned me today when I was trying to debug some issues, and there is an existing issue for this https://github.com/blockstack/docs/issues/880.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

As far as I know no

## Are documentation updates required?
No

## Testing information

To test go through the stacks cli page and verify the references are replaced.

## Checklist
- [ ] Code is commented where needed
- [ ] Unit test coverage for new or modified code paths
- [ ] `npm run test` passes
- [ ] Changelog is updated
- [ ] Tag 1 of @person1 or @person2


agraebe:
@yknl could you review?

friedger:
This should be fixed in https://github.com/blockstack/stacks.js/blob/master/packages/cli/src/argparse.ts

friedger:
There are two commands available for the command line, `stx` and `stacks`. In the cli wallet quickstart tutorial, we used `stacks`. Is that the way to go for all documentation, @agraebe ?

yknl:
The `cli_reference.json` is automatically generated. As @friedger pointed out it should be updated in `argparse.ts`

> There are two commands available for the command line, stx and stacks. In the cli wallet quickstart tutorial, we used stx. Is that the way to go for all documentation

We should use `stx` since it's shorter.

agraebe:
can this be closed in favor of the other merged PR?

alexkeating:
I think the cli-reference.json still needs to be regenerated and the README needs to be updated with the stx command.

agraebe:
@yknl can you check the readme updates required here? briefly looking at it and it seems to essentially link to: https://docs.blockstack.org/references/blockstack-cli. This page is using the old alias.

yknl:
> @yknl can you check the readme updates required here? briefly looking at it and it seems to essentially link to: https://docs.blockstack.org/references/blockstack-cli. This page is using the old alias.

Can you clarify? This PR should be closed as https://github.com/blockstack/docs/pull/946 supercedes it and has been merged.

agraebe:
ah, sorry! you're right.

PR #936, opened 2020-11-30
Update mining instructions for Xenon.

diwakergupta:


src/pages/mining.md
```diff
@@ -23,6 +23,33 @@ If you want to learn more about the technical details of minig, please review th
...
+txindex=0
+listen=1
+rpcserialversion=0
+maxorphantx=1
+banscore=1
+
+[test]
+bind=0.0.0.0:18443
+rpcbind=0.0.0.0:18444
+rpcport=18443
```
CharlieC3, line 48: We're now using the standard BTC testnet ports for our deployment. Since we're allowing users running a follower to connect to our BTC node's p2p port, we should ensure users running a miner use the same BTC ports for uniformity.
```suggestion
bind=0.0.0.0:18333
rpcbind=0.0.0.0:18332
rpcport=18332
```

src/pages/mining.md
```diff
@@ -59,18 +86,20 @@ Paste in the following configuration:
...
-#commit_anchor_block_within = 10000
+mode = "xenon" # if connecting to Krypton, set this to "krypton"
+# To mine on Xenon, you need to run bitcoind locally (details below)
+# For Krypton, set peer_host to `bitcoind.krypton.blockstack.org` and
+# omit `username` and `password`
+peer_host = "127.0.0.1"
+username = "your-bitcoind-username"
+password = "your-bitcoind-password"
 rpc_port = 18443
 peer_port = 18444
```
CharlieC3, line 104: Ties in with my above suggestion
```suggestion
rpc_port = 18332
peer_port = 18333
```

CharlieC3 (changes requested):
Are updates for running a Xenon _follower_ coming in a separate PR? I may be able to lend a hand if needed.

src/pages/mining.md
diwakergupta: Sure, happy to change it. I was trying to minimize diff from the Krypton config.

diwakergupta:
> Are updates for running a Xenon _follower_ coming in a separate PR? I may be able to lend a hand if needed.

Sure, go for it @CharlieC3 !

CharlieC3:
@diwakergupta
* Updated some of the other configs found on the `miner` page to more closely resemble the one you edited
* Updated the docker command to use the `--network` flag to ensure the stacks miner can communicate with the user's local BTC node.
* Updated various references found in the stacks-node cli page

CharlieC3:
@lgalabru Correct me if I'm wrong, but doesn't a BTC command need to be run against the user's local bitcoind node before the stacks miner can use their private key?

diwakergupta:
> @lgalabru Correct me if I'm wrong, but doesn't a BTC command need to be run against the user's local bitcoind node before the stacks miner can use their private key?

Yeah, it's `importaddress`. Something like:

`bitcoin-cli  -rpcport=18332 -rpcuser=your-user -rpcpassword=your-password importaddress <your-btc-addr>`

@CharlieC3 do you want to update the PR to include that and then let's merge this in?

CharlieC3:
Added the command to the docs, not sure if the ordering is correct though. I assume you need to import the BTC address before hitting the BTC testnet faucet.

@diwakergupta or @lgalabru Can you please verify?
https://blockstack-docs-azaxwc94v.vercel.app/mining#running-a-miner

CharlieC3 (approved):

diwakergupta:
> Added the command to the docs, not sure if the ordering is correct though. I assume you need to import the BTC address before hitting the BTC testnet faucet.

No, you can hit any tBTC faucet anytime. All the `importaddress` does is tell bitcoind to start monitoring UTXOs for that address. You do need to hit the faucet and run `importaddress` before starting `stacks-node`, or it won't find UTXOs for that address.

Thanks for shepherding this through @CharlieC3 !

CharlieC3:
Merged, thanks for the info @diwakergupta !

PR #939, opened 2020-12-01
Update todos app tutorial

markmhx:
This PR updates the todos app tutorial in line with its underlying updates with https://github.com/blockstack/todos/pull/57, namely fixing authentication guidance, while also making a bunch of other general improvements (e.g. to language and screenshots).

Resolves https://github.com/blockstack/docs/issues/911


package-lock.json
```diff
@@ -0,0 +1,13271 @@
+{
```
agraebe, line 1: @aulneau my assumption is we should not commit this change as part of a content update. correct?

src/pages/authentication/overview.md
```diff
@@ -35,9 +35,9 @@ Scopes define the permissions requested by an app for granting during authentica
 Apps may request any of the following scopes:
 
 | Scope          | Definition                                                                           |
-| -------------- | ------------------------------------------------------------------------------------ |
+| -------------- | ------------------------------------------------------------------------------------ | --- |
```
agraebe, line 38: this creates a third empty column, can we remove that?

src/pages/build-an-app.md
```diff
@@ -1,14 +1,14 @@
 ---
-title: Building decentralized apps
-description: Overview and guides for getting started building decentralized applications.
+title: Building apps with Stacks
+description: Resources for getting started with building apps on the network
```
agraebe, line 3: `the network` > `Stacks 2.0`?

src/pages/authentication/building-todo-app.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Building a Todo app
-description: Learn how to integrate authentication and data storage.
+title: Review Todos app
```
agraebe, line 2: I'm a bit puzzled by this - why would a developer come here to review an app? I think the goal should be to build/deploy an app. Could we redesign this tutorial to include active steps / changes devs would have to make throughout the tutorial?

I'd suggest we fork a branch of the todo app that has comments in several places where changes should be made to make the app work. The branch would be like a boilerplate, ready to be completed during the tutorial

@hstove - thoughts?

agraebe (changes requested):
let's align more high level on the goal/purpose of the tutorial. Reviewing existing code will not create a good learning journey IMHO

src/pages/authentication/overview.md
aulneau: If you rebase from master this will be fixed

src/pages/build-an-app.md
aulneau: yes, agreed `Stacks 2.0` is better

src/pages/authentication/building-todo-app.md
markmhx: I'd prefer to get these fixes in place before we open up the question of whether this "review" type of material vs. "build" is better (and reworking the entire structure if we pursue the latter, which I agree could have greater merit). Currently this is the only way in which our docs exhibit Stacks.js-based authentication for Connect clearly, so it pains me to leave them essentially broken and misleading while we work out higher level questions.

Also keep in mind that after getting this merged, it's top priority for me to put in place a build-like tutorial sequence for auth, storage and transaction signing. That fresh effort may eventually replace this todos list app entirely, though I do think code samples with guidance have value.

src/pages/build-an-app.md
markmhx: Is "Stacks 2.0" really a version we want to enshrine in our materials generally going forward? It seems a bit weird to put "2.0" front and center for those who aren't tracking the 1.0 vs. 2.0. difference / chain launch specifically. It'll become even more outdated as the months go on.

A newcomer developer is likely to look at "2.0" and go "huh? why 2.0?", since that version only really matters to those who're helping to prepare or track the launch of Stacks 2.0 specifically.

I could certainly see us phrasing this as "Resources for getting started with building apps on Stacks" without a version, though my hope here was to help teach that "Stacks" == "network" with the relative use of the title and description.

src/pages/authentication/overview.md
markmhx: Agreed, I'll remove via rebase

src/pages/authentication/building-todo-app.md
agraebe: I'm not sure I get it. Why would someone look for a review of an app in the journey of learning about a library?

I think there's a low hanging fruit to introduce some changes to make this a more meaningful tutorial, without reworking everything. The content in here can still be used mostly.

Just pushing this out at this stage won't be impactful I believe. It'll allow us to say there's content but not that there's useful content. I'd suggest we make small corrections to make this a real tutorial.

src/pages/build-an-app.md
agraebe: Maybe we just say "on the Stacks blockchain"?

src/pages/authentication/building-todo-app.md
markmhx: Someone may very well want to look at how a simple app is built with a certain technology or library IMO. This is personally how I like to learn new tech, fwiw. But I also agree it's not necessarily the only or best way.

I think what you're proposing is that we basically make the user copy and paste snippets into a more partial todos code base, which I'm concerned would just make busy work for them (since copy and pasting doesn't really make them think). We could make it more engaging and complicated, and I do think that's worth exploring, but that's going to take time. So in the meantime, I don't see why it hurts getting this merged, feeling confident our docs aren't broken, then starting that effort (immediately)?

src/pages/build-an-app.md
markmhx: That works for me. I'd heard before @cuevasm comment that we'd do well to use "network" going forward vs. "platform" at least. But perhaps "Stacks blockchain" is preferred to "Stacks network" these days?

cuevasm: +1 try to avoid platform, it has centralized connotations. Network or blockchain is preferable!

src/pages/authentication/building-todo-app.md
agraebe: We had a convo in discord and I'll put a tl;dr in here: if you believe the content (as-is) will have the impact you're looking for on the product side, I'm fine merging it. 

Outside of this particular instance, I would urge us not to get comfortable with a "review" type of tutorial in other places.

agraebe (approved):

src/pages/authentication/building-todo-app.md
aulneau: I would agree that this tutorial is actually a review of how to build an app vs a step by step implantation from scratch.

package-lock.json
aulneau: This file should not be here, you are using npm where you should use `yarn`

src/pages/authentication/building-todo-app.md
markmhx: Thanks 🙏  I agree, let's shoot higher than "code review" for content structure going forward.

package-lock.json
markmhx: Oops, got it. `npm` is a habit for me. Should we remove support from the repo if `yarn` is preferred? 

I'll remove this file meanwhile.

kyranjamie:
@markmhx think you need to rebase this PR, there're other commits in the branch

aulneau:
closing in favor of this rebased version https://github.com/blockstack/docs/pull/945

src/pages/build-an-app.md
```diff
@@ -1,14 +1,14 @@
 ---
-title: Building decentralized apps
-description: Overview and guides for getting started building decentralized applications.
+title: Building apps with Stacks
+description: Resources for getting started with building apps on the Stacks blockchain
```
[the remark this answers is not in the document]
markmhx: Cool call, I've updated to do "Resources for building apps on the Stacks blockchain" on the new branch / PR linked below.

PR #940, opened 2020-12-01
Adds untested link fix; clarifies language for similar link.

jbheron:
## Broken Docs Link Fix

@agraebe, could you take a look at this suggested link fix? Not sure if this is the correct approach, or what I may have missed. 🙏 

I was looking at the docs tooling and organization and found a broken link. Thought I'd make a quick PR to fix it. I reviewed the  contributor README and the navigation.yaml, but didn't take the time to test by installing npm, yarn, and other dependencies, so this is just a guess to the correct syntax for the link.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] API reference/documentation update
- [x] Other: link fix

As mentioned, **I did not test this change**, so _maybe_?

## Checklist
- [x] Code is commented where needed
- [ ] Unit test coverage for new or modified code paths
- [ ] `npm run test` passes
- [ ] Changelog is updated
- [x] Tag 1 of @person1 or @person2

### DIFF
--- a/src/pages/storage-hubs/amazon-ec2-deploy.md
+++ b/src/pages/storage-hubs/amazon-ec2-deploy.md
@@ -13,7 +13,7 @@ disk for file storage.
 
 -> This documentation is appropriate for advanced power users who are familiar with command line tools, `ssh`, and basic editing configuration files.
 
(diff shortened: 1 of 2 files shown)


agraebe (approved):
🙏 thanks!

Issue #952, opened 2020-12-07
Improve CLI references


agraebe:
The CLI references are: 

- [x] using the `blockstack-cli` name in the URL
- [x] not linked to from within other guides/tutorials
- not well-formatted
- lack categorization based on chapters

PR #953, opened 2020-12-08
Fixing CLI references

agraebe:
Partially closing #952

### DIFF
--- a/src/pages/references/stacks-cli.md
+++ b/src/pages/references/stacks-cli.md
@@ -28,30 +28,6 @@ You must have [Node.js](https://nodejs.org/en/download/) v12 or higher (v14 reco
 
 To install the command line, do the following:
 
-### Step 1: [Download or `git clone` the command line repository code](https://github.com/blockstack/stacks.js/tree/master/packages/cli)
-
-Downloading or cloning the repo creates a `cli-blockstack` repository on your system.
-
-### Step 2: Change directory into the `cli-blockstack` directory
-
-```bash
-cd cli-blockstack
-```
-
-### Step 3: Install the dependencies with `npm`
-
-```bash
-npm install
-```
-
-### Step 4: Build the command line command
-
-```bash
-npm run build
-```
-
-### Step 5: Link the command
-
 ```bash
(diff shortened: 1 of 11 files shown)


src/pages/references/stacks-cli.md
```diff
@@ -27,6 +28,30 @@ You must have [Node.js](https://nodejs.org/en/download/) v12 or higher (v14 reco
 
 To install the command line, do the following:
 
+### Step 1: [Download or `git clone` the command line repository code](https://github.com/blockstack/stacks.js/tree/master/packages/cli)
```
yknl, line 31: I think we should just delete the clone repo instructions and use `npm install -g @stacks/cli` here.

src/pages/understand-stacks/install-api.md
```diff
@@ -24,8 +24,8 @@ $ cd blockstack-core/
 $ git checkout stacks-1.0
 $ pip install .
 $ pip install -r api/requirements.txt
-$ blockstack setup_wallet
-$ blockstack api start
+$ stx setup_wallet
```
yknl, line 27: This is very outdated documentation and no longer relevant. We should just delete the whole thing.

yknl (approved):

src/pages/understand-stacks/install-api.md
agraebe: can we delete the whole file?

src/pages/references/stacks-cli.md
agraebe: good point! I didn't look closely enough. will us npm instead

src/pages/understand-stacks/install-api.md
yknl: These instructions are for the 1.0 blockstack-core. Either we remove the whole page, or add a notice at the top that this is out of date and remove it after 2.0 mainnet.

agraebe: added a callout

aulneau:
Please rebase and push to fix the build issue.

Issue #956, opened 2020-10-09
Add hardware requirements for running a node 


GinaAbrams:
Hello, wondering if we can add the hardware requirements for running a mainnet stacking node to the readme of this repo?

Let me know if this is a better question for docs. Thanks!

PR #957, opened 2020-12-08
add mining hardware info

agraebe:
Fixing #956

### DIFF
--- a/src/pages/understand-stacks/running-testnet-node.md
+++ b/src/pages/understand-stacks/running-testnet-node.md
@@ -23,6 +23,29 @@ This tutorial will walk you through the following steps:
 
 ## Requirements
 
+In order to run a node, some software and hardware requirements need to be considered.
+
+### Hardware
+
+Running a node has no specialized hardware requirements. People were successful at running a node on Raspberry Pis, for instance.
+Minimum requirements are moving targets due to the nature of the project and some factors should be considered:
+
+- compiling node sources locally requires processing and storage resources
+- replicating the chainstate requires significant memory allocation (1.3+ GB)
+- chainstate size grows over time, increasing the memory requirements
+
+With these considerations in mind, we suggest hardware based on the general-purpose specification provided by GCP, the [E2 machine standard 2](https://cloud.google.com/compute/docs/machine-types#general_purpose):
+
+- 2 vCPUs
+- 8 GB memory
+- ~50-GB disk (preferably SSDs)
+
+It is also recommended to run the node with a publicly routable IP, that way other peers in the network will be able to connect to it.
+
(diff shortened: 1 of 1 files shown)


src/pages/understand-stacks/running-testnet-node.md
```diff
@@ -23,6 +23,29 @@ This tutorial will walk you through the following steps:
...
+In order to run a node, some software and hardware requirements need to be considered.
+
+### Hardware
+
+Running a node has no specialized hardware requirements. People were successful at running a node on Raspberry Pis, for instance.
+Minimum requirements are moving targets due to the nature of the project and some factors should be considered:
+
+- compiling node sources locally requires processing and storage resources
+- replicating the chainstate requires significant memory allocation (1,3+ GB)
+- chainstate size grows over time, increasing the memory requirements
```
diwakergupta, line 35: I think there's some confusion here and in the prior point:

* as the chain grows, the *on-disk state* will grow over time. Memory requirements are relatively static (not a function of chain-length, mostly the mempool).
* "Replicating the chainstate requires significant memory allocation" --  I think what you may referring to here is memory requirements for *ingesting the Stacks 1.0 state* -- those are indeed higher (~2-GB RAM) but only relevant to nodes that are upgrading from Stacks 1.0 or participating in Stacks 2.0 launch. If you spin up a node post launch, the memory requirements to sync w/ the network would be similar to what it was in Krypton or Xenon, which is to say, not a lot.
- "1,3+ GB" -- typo, perhaps you meant "1-3 GB"?

src/pages/understand-stacks/running-testnet-node.md
```diff
@@ -23,6 +23,29 @@ This tutorial will walk you through the following steps:
...
+
+With these considerations in mind, we suggest hardware based on a general-purpose specification provided by GCP, the [E2 machine standard 2](https://cloud.google.com/compute/docs/machine-types#general_purpose):
```
diwakergupta, line 37: Perhaps add an equivalent AWS instance as well for completeness? It's not like we're recommending people use GCP or any specific vendor.

diwakergupta (approved):
Some suggestions on clarifying the language, otherwise LGTM.

src/pages/understand-stacks/running-testnet-node.md
agraebe: I'll remove the memory allocation point (2nd in the list) and alter the chain growth point. thanks for the feedback!

src/pages/understand-stacks/running-testnet-node.md
agraebe: great point, adding that as well

PR #964, opened 2020-12-11
feat: full-width docs

agraebe:
Fixes blockstack/docs#947

### DIFF
--- a/src/common/constants.ts
+++ b/src/common/constants.ts
@@ -1,5 +1,6 @@
 export const SIDEBAR_WIDTH = 208;
 export const TOC_WIDTH = 208;
+export const CONTENT_MAX_WIDTH = 1175;
 
 export const SHIKI_THEME = 'Material-Theme-Default';
 export const THEME_STORAGE_KEY = 'theme';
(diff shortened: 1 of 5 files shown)


kyranjamie (approved):
Big improvement, thanks @agraebe

aulneau:
I'm generally in favor of this change, but I think we should hear from @jasperjansz before merging this in. Additionally, I wonder if we should have the TOC on the right, all the way to the right?

agraebe:
> I'm generally in favor of this change, but I think we should hear from @jasperjansz before merging this in. Additionally, I wonder if we should have the TOC on the right, all the way to the right?

Agreed, been waiting to hear from @jasperjansz but couldn't get anyone's eyes on this for a week. I had a version of this on the very right and it didn't look great - felt like it was disconnected from the content. I looked at Stripe docs and they seem to have the same layout (which shouldn't be a strong argument, but it's _a signal_)

jasperjansz:
Im fine with this change.

If you want to make it read easier, you can decrease the width of the center column so you'll have less words per line. Currently you're close to 130 while 80-90 is optimal for reading for prolonged periods.

agraebe:
please have another look at the deployed version. I set the width to 860px to optimize the reading experience as per @jasperjansz's advice

Issue #972, opened 2020-12-15
URL checker


agraebe:
To avoid broken links, we should check all URLs inside our markdown files

PR #973, opened 2020-12-15
feat: add URL checker and fix broken URLs

agraebe:
Fixing #972


aulneau (changes requested):
Nice, this is a different tool than I would have used, but seems to work. Please remove `package-lock.json`

mlc_config.json
```diff
@@ -0,0 +1,23 @@
...
+    }
+  ],
+  "replacementPatterns": [
+    {
+      "pattern": "^/public/images",
+      "replacement": "https://docs.blockstack.org/images"
+    },
+    {
+      "pattern": "^/",
+      "replacement": "https://docs.blockstack.org/"
```
CharlieC3, line 17: Same as other comment

mlc_config.json
```diff
@@ -0,0 +1,23 @@
...
+      "pattern": "^http://localhost"
+    },
+    {
+      "pattern": "^https://banter.pub"
+    }
+  ],
+  "replacementPatterns": [
+    {
+      "pattern": "^/public/images",
+      "replacement": "https://docs.blockstack.org/images"
```
CharlieC3, line 13: With this configuration, does that mean the tool is going to look for references to an image which matches the regex pattern `^/public/images`, and replace it with `https://docs.blockstack.org/images`? If so, I'm not sure this is very portable, since it will break if/when we update our site to `hiro.so` in the future. What's wrong with using `^/public/images`?

CharlieC3 (commented):
Looks good, just a couple comments.

mlc_config.json
agraebe: that's right, it replaces the relative image paths with absolute paths that are accessible once deployed. This is done so that images can be checked. otherwise, all relative paths will throw an error.

It is not very portable. Any suggestions to improve that? I would guess that we'd search for and replace "docs.blockstack.org" anyways inside the entire repo if/when a domain change would happen. In that case, it wouldn't be a big deal?

mlc_config.json
agraebe: replacing relative paths with accessible paths of deployed resources. see the other comment for details :)

mlc_config.json
CharlieC3: > This is done so that images can be checked

I could be wrong in my thinking, but I'm not sure a change like this would achieve this goal. If someone were to create a new PR which adds a brand new image to their branch, if they try to reference the image using the full production URL (`https://docs.blockstack.org/images`), it would error in testing because the image doesn't exist yet in prod. They would need to use a relative path for the app to use the new image from `https://blockstack-docs-jyzf58fjr.vercel.app/images` and not `https://docs.blockstack.org/images` because it doesn't exist there yet until the PR is merged and the update is auto-deployed.

It's the same reason why we use relative paths when linking one docs page in another docs page. Because if it's a new docs page we're linking, it wouldn't resolve to a valid page in `https://docs.blockstack.org`.

Sorry if I'm not very clear on this, happy to show you what I mean in Zoom if needed.

agraebe: you're correct, this is one of the edge cases that would be broken. @aulneau pointed that out as well today. I think there are three approaches to look at here:

- n.1: run this check on master [as a scheduled task](https://github.com/gaurav-nelson/github-action-markdown-link-check#scheduled-runs), but do not make it a mandatory check for PRs to be merged. When this check is executed on deployed code, the edge case isn't relevant. BTW, this check is _not_ proposed to be run at commit/PR time. I just added an npm task, so it can be run manually
- n.2: run a github action, similar to this check on every PR (https://github.com/blockstackpbc/devops/issues/533). In this case, we're presented with the issue you pointed out. We could use the (potentially wrong) annotations as a mechanism to reinforce manual URL checks during PRs. Or, we just don't do these checks and rely on the scheduled runs 
- n.3: run a github action but with a different tool that would have awareness of the project folder structure and could figure out if a PR includes the images/files in reference. @aulneau mentioned he was aware of such a tool. I'd assume this would require some more thinking/work

tl;dr: I think periodic URL checks are better than no checks, which is where we are right now. I'd propose we try n.1 and iterate with an eye on n.3. wdyt @CharlieC3 ?

CharlieC3: I think I'm just hung up on _why_ using relative references here is a bad idea. Using references and limiting hard-coded values is the paradigm most elsewhere, why do we _want_ to hard-code this? You mentioned the following:

> This is done so that images can be checked. otherwise, all relative paths will throw an error.

What is doing these checks? Where is this error appearing? The images load for me just fine in the prod docs site and in PRs using referenced images.

agraebe: the library used (markdown link checker) simply makes HTTP GET calls to a URL provided. The tool extracts all URLs from given markdown files. 

Our markdown files, by convention, use only relative paths. Without a replacement of the relative path to an absolute path, the HTTP call (coming from the library) would return a 404, so every link to another page inside the docs would throw an error. Same for images (but with a different absolute path generated). 

I wish the library would have awareness of the project file structure and could figure out relative paths, but it doesn't. So the workaround seems to be to create "replacement" instructions using regex for every relative path.

CharlieC3: Ah ok, thanks for the clarification. While I do like the idea of checking markdown links, I don't think we should implement this specific tool in our workflow for a couple reasons:

1. The tool is forcing us to deliberately introduce a poor practice, especially because we're already able to identify a very likely point in time when this will break for us despite best efforts to prevent it - when we switch to `hiro.so`.
2. We'll have a laundry list of things to change during the switch to `hiro,so` to minimize breakages, and I would rather avoid adding another potential weak link on top of that.

Maybe the tool @aulneau mentioned would be able to give us the best of both worlds? @aulneau do you remember the name of it?

agraebe: I do want to note that while I do agree that there's a better approach to be found, IMHO the downside of broken links is much higher than the downside of changing one additional hardcoded URL in the docs project. There will be many changes of URLs in the coming weeks/months on the docs and the trade-off of updating a hard-coded URL (which is only one occurrence, there's already precedent for the hardcoded URL inside the header, the fathom integration, and even in some links as I just realized).

I would argue this is a good first step towards more accurate documentation and shouldn't be minimized for our operational reasons. I'm afraid that, with very limited resources focusing on docs, looking for another tool and building out another PR is likely so much effort, that we'll not even get to _any_ link checker in the next weeks/months.

If the concern is not justified and @aulneau can help, that's fine. But let's not be the perfect a roadblock to the good.

CharlieC3: Agreed the tool does provide some value, I'm just concerned that it's creating new holes while we plug up old ones. I think it's also worth noting the tool won't be able to detect any broken links until they're already deployed to production due to the reason I mentioned in my second comment above. So while it certainly can help us, it would only be able to report any broken links after they're already in place, and would likely result in someone having to quickly push out a fix.

This is still much better than not having _anything_ checking the links like where we are today, but just want to set expectations that we won't be able to _prevent_ link breakages with this tool it seems, just detect where they are in production.

I think the points you raised are important @agraebe, I'm good to move forward with this, but would like us to investigate the other tool @aulneau has in mind at a later date.

CharlieC3 (approved):
:shipit:

Issue #912, opened 2020-11-16
Review vale for style guide checks


agraebe:
Review this project: https://github.com/errata-ai/vale. This could help unify the writing style across docs

agraebe:
A local setup of the project with the style-guide from #850 works quite well. What needs to be figured out is this:

How can we incrementally update the style issues? There are hundreds that would require hours of work. Maybe we could start with checks for PRs for now.

@CharlieC3 could you help by reviewingg https://github.com/errata-ai/vale-action ?

PR #974, opened 2020-12-15
feat: add language style checker, fix style errors

agraebe:
Fixing #912


src/pages/data-storage/storage-guide.md
```diff
@@ -26,7 +26,7 @@ const options: PutFileOptions = {
   encrypt: false,
 };
 userSession.putFile('hello.txt', 'hello world!', options).then(() => {
-  // hello.txt exists now, and has the contents "hello world!".
+  // hello.txt exists now, and has the contents "hello world!."
```
friedger, line 29: That looks wrong

src/pages/ecosystem/contributing.md
```diff
@@ -46,8 +46,8 @@ When workin locally with the site, a few things are needed:
 
 All of the code for this site is open source, located at the [GitHub repository here](https://github.com/blockstack/docs).
 Before you start editing anything, you will need to fork the repo so that you can have your own copy of the code under
-your GitHub profile. On the [repository's page](https://github.com/blockstack/docs), you should be able to
-see a button in the upper right of the screen that says "Fork". [You can read about Forking here.](https://docs.github.com/en/github/getting-started-with-github/fork-a-repo)
+your GitHub profile. On the [repository's page](https://github.com/blockstack/docs.blockstack), you should be able to
+see a button in the upper right of the screen that says "Fork." [You can read about Forking here.](https://docs.github.com/en/github/getting-started-with-github/fork-a-repo)
```
friedger, line 50: Wrong change

src/pages/ecosystem/contributing.md
```diff
@@ -219,7 +227,7 @@ Which will render:
 
 (define-public (test-emit-event)
     (begin
-        (print "Event! Hello world")
+        (print "Event. Hello world")
```
friedger, line 230: This could be a!

src/pages/write-smart-contracts/principals.md
```diff
@@ -61,15 +61,15 @@ example below.
...
+contract _and_ the contract's name, for example:
 
 ```clarity
 'SZ2J6ZY48GV1EZ5V2V5RB9MP66SW86PYKKQ9H6DPR.contract-name
 ```
 
 For convenience, smart contracts may write a contract's identifier in the
 form `.contract-name`. This will be expanded by the Clarity interpreter into
-a fully-qualified contract identifier that corresponds to the same
+a fully qualified contract identifier that corresponds to the same
```
friedger, line 72: Should be fully-qualified

friedger (commented):
I like the style guide! It could be adopted to stacks. At least layout how to add stacks specific styles in the files

Some ! could have stayed

.vale/styles/Google/Acronyms.yml
```diff
@@ -0,0 +1,64 @@
+extends: conditional
+message: "Spell out '%s', if it's unfamiliar to the audience."
+link: 'https://developers.google.com/style/abbreviations'
+level: suggestion
+ignorecase: false
+# Ensures that the existence of 'first' implies the existence of 'second'.
+first: '\b([A-Z]{3,5})\b'
+second: '(?:\b[A-Z][a-z]+ )+\(([A-Z]{3,5})\)'
+# ... with the exception of these:
+exceptions:
```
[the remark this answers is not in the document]
agraebe: great catch. the tool actually uses another list for exceptions, outside of the style. I'm inclined to leave the Google style untouched and add our exception in the other list. That way, we could update the styles whenever needed. It also makes using the GH action easier (the GH action requires the release files of a style guide as a ZIP)

.vale/styles/Google/Headings.yml
```diff
@@ -0,0 +1,26 @@
+extends: capitalization
+message: "'%s' should use sentence-style capitalization."
+link: 'https://developers.google.com/style/capitalization#capitalization-in-titles-and-headings'
+level: warning
+scope: heading
+match: $sentence
+exceptions:
+  - Azure
```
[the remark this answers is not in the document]
agraebe: same as above, added it to the accept list to make exceptions for these words

src/pages/data-storage/storage-guide.md
agraebe: fixed

src/pages/ecosystem/contributing.md
agraebe: this is an interesting one because it should by default skip all sample codes. Must be because we have a docs and clarity specific annotation there: "```clarity highlight=1,4-6,13-17,28-32".

As opposed to spending more time, I'd rather just change up the print command to keep things simple for now

src/pages/write-smart-contracts/principals.md
agraebe: this is against the rules and I think for good reasons. [FQDNs](https://en.wikipedia.org/wiki/Fully_qualified_domain_name) don't use the hyphen.

Ultimately, if we believe exceptions like this should be ok, we should start working on our own styleguide. The complexity would blow up quite quickly. I'd rather just use a default for now and make our life easier

src/pages/ecosystem/contributing.md
agraebe: this is actually according to the rules. if the quote is at the end of the sentence, the comma or dot should be inside the quote. Another "controversial" guideline. I'd just follow it for now and work on our own guidelines later

PR #976, opened 2020-12-16
Add UTF-8 string example

alexkeating:
## Description

Add UTF-8 string example based on https://github.com/blockstack/docs/issues/944.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Other

### DIFF
--- a/src/pages/references/language-types.md
+++ b/src/pages/references/language-types.md
@@ -27,3 +27,4 @@ The type system contains the following types:
 - `bool` := boolean value (`true` or `false`)
 - `int` := signed 128-bit integer
 - `uint` := unsigned 128-bit integer
+- `u"string"` := UTF-8 string (UTF-8 bytestring `u"\u{string}"`)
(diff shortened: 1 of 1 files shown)


src/pages/references/language-types.md
```diff
@@ -27,3 +27,4 @@ The type system contains the following types:
 - `bool` := boolean value (`true` or `false`)
 - `int` := signed 128-bit integer
 - `uint` := unsigned 128-bit integer
+- `u"string"` := UTF-8 string (UTF-8 bytestring `u"\u{string}"`)
```
friedger, line 30: It should be something like the one below and in line 24:
On the left there is the type name on the right a description with example

(string-utf8 max-len) := UTF-8 string of maximum length max-len (`u"An emoji \u{1F600} as a utf8 string"`)

friedger (changes requested):
You should add the example at the existing type definition

friedger (approved):
Looks good

PR #977, opened 2020-12-16
s/mstx_balance/ustx_balance/ in stacks-node config.

diwakergupta:


CharlieC3 (approved):
@diwakergupta Will stage this change in k8s to be used the next time we upgrade Xenon.

PR #982, opened 2020-12-17
use xenon.blockstack.org as bootstrap node

friedger:
## Description
This PR changes the bootstrap node from testnet-miner.blockstack.org to xenon.blockstack.org

- [x] @agraebe


CharlieC3 (approved):
:shipit:

CharlieC3:
@friedger Just need to fix the linting issue then should be good to go!

PR #985, opened 2020-12-17
feat: add github action for vale check

agraebe:
There is a Vale action and config files in the docs repo, but we don't have a check that runs on PRs. This change adds a Github action to compare changed files to the styleguide and point out suggestions/errors.

irt https://github.com/blockstackpbc/devops/issues/527

this also considers a known Vale issue: https://github.com/errata-ai/vale-action/issues/27

### DIFF
--- a/.github/workflows/stylecheck-prs.yml
+++ b/.github/workflows/stylecheck-prs.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           token: ${{ github.token }}
       - name: Run vale
-        uses: errata-ai/vale-action@v1.3.0
+        uses: errata-ai/vale-action@master
         with:
           styles: https://github.com/errata-ai/Google/releases/latest/download/Google.zip
           config: https://raw.githubusercontent.com/blockstack/docs/master/.vale/vale.ini
--- a/.github/workflows/stylecheck-prs.yml
+++ b/.github/workflows/stylecheck-prs.yml
@@ -1,0 +1,25 @@
+name: Stylecheck
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  stylecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Get Changed Files
+        id: get_changed_files
+        uses: lots0logs/gh-action-get-changed-files@2.1.4
+        with:
+          token: ${{ github.token }}
+      - name: Run vale
+        uses: errata-ai/vale-action@v1.3.0
+        with:
+          styles: https://github.com/errata-ai/Google/releases/latest/download/Google.zip
+          config: https://raw.githubusercontent.com/blockstack/docs/master/.vale/vale.ini
+          files: "${{ steps.get_changed_files.outputs.all }}"
+          onlyAnnotateModifiedLines: true
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
--- a/.github/workflows/stylecheck-prs.yml
+++ b/.github/workflows/stylecheck-prs.yml
@@ -2,6 +2,8 @@ name: stylecheck
 on:
   pull_request:
     branches: [master]
+    paths:
+      - '**.md'
 
 jobs:
   stylecheck:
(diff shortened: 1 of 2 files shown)


.github/workflows/stylecheck-prs.yml
```diff
@@ -0,0 +1,25 @@
...
+  stylecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: get Changed Files
+        id: get_changed_files
+        uses: lots0logs/gh-action-get-changed-files@2.1.4
+        with:
+          token: ${{ github.token }}
```
CharlieC3, line 16: Based on the docs, GH uses `${{ secrets.GITHUB_TOKEN }}` everywhere they need it. Would you mind changing this and the other usage of it to `${{ secrets.GITHUB_TOKEN }}` ?

CharlieC3 (changes requested):
One small change to match the GH docs for readability.

.github/workflows/stylecheck-prs.yml
agraebe: sure!

.github/workflows/stylecheck-prs.yml
```diff
@@ -0,0 +1,25 @@
...
+        id: get_changed_files
+        uses: lots0logs/gh-action-get-changed-files@2.1.4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: run vale
+        uses: errata-ai/vale-action@master
+        with:
+          styles: https://github.com/errata-ai/Google/releases/latest/download/Google.zip
+          config: https://raw.githubusercontent.com/blockstack/docs/master/.vale/vale.ini
+          files: "${{ steps.get_changed_files.outputs.all }}"
```
CharlieC3, line 22: Is there any way we can exclude some files from being analyzed? For example, vale should never analyze files changes in `.github`. Maybe it would be easier to just specify certain dirs like `src` and `public` (and any others)?

.github/workflows/stylecheck-prs.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: get changed Files
+        id: get_changed_files
+        uses: lots0logs/gh-action-get-changed-files@2.1.4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: run vale
+        uses: errata-ai/vale-action@master
```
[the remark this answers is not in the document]
agraebe: I used that before but the master version is the only one that enables limiting checks only to the lines that were changed. if we bump to the 1.3 version, the checks will apply to the entire file that was changed, which is very likely overkill

.github/workflows/stylecheck-prs.yml
agraebe: good point... uhm let me see what is possible here

.github/workflows/stylecheck-prs.yml
```diff
@@ -0,0 +1,25 @@
+name: stylecheck
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  stylecheck:
+    runs-on: ubuntu-latest
+    steps:
```
CharlieC3, line 9: This action's GH page says the following:
> Due to the current token permissions, this Action CAN NOT post annotations to PRs from forked repositories.

The following will restrict this job to run only on non-forked PRs.

```suggestion
    if: github.event.pull_request.head.repo.full_name == github.repository
    steps:
```

.github/workflows/stylecheck-prs.yml
agraebe: looks like this might be an option: https://docs.errata.ai/vale/config#format-associations. would still analyze the files but ignore them it looks like

.github/workflows/stylecheck-prs.yml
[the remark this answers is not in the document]
CharlieC3: Gotcha, seems they haven't released since August, guess that's why. Good enough for me!

PR #995, opened 2020-12-28
docs: update stacking delegation info

agraebe:
Fixes blockstack/docs#994

### DIFF
--- a/src/pages/understand-stacks/integrate-stacking.md
+++ b/src/pages/understand-stacks/integrate-stacking.md
@@ -331,8 +331,10 @@ Stacking delegation requires you to implement a different flow. Until the provid
 
 Below is a description of contract calls required to integrate delegation:
 
-1. Account holders call the [`delegate-stx`](/references/stacking-contract#delegate-stx) method and provide the delegator address, the amount accessible to the delegator, the address rewards should be sent to, and the burn height at which the delegation relationship should be terminated
-2. For each delegation relationship that was created, the delegator calls the [`delegator-stack-stx`](/references/stacking-contract#delegate-stack-stx) method to lock up the STX token from the account holder. This method must be called until the delegator locked up enough STX tokens required to participate in Stacking
+1. Account holders call the [`delegate-stx`](/references/stacking-contract#delegate-stx) method and provide the delegator address, the amount accessible to the delegator, the address rewards should be send to, and the burn height at which the delegation relationship should be terminated. This needs to happen only once. The account holder can specify a delegation amount that is higher that their balance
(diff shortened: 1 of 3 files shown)


src/pages/understand-stacks/network.md
```diff
@@ -100,7 +100,7 @@ Sample response for a successful call:
 }
 ```
 
--> To set the function call arguments and read the result, [Clarity values](http://localhost:3000/understand-stacks/transactions#clarity-value-types) need to be serialized into a hexadecimal string. The [Stacks Transactions JS](https://github.com/blockstack/stacks.js/tree/master/packages/transactions) library supports these operations
+-> To set the function call arguments and read the result, [Clarity values](/understand-stacks/transactions#clarity-value-types) need to be serialized into a hexadecimal string. The [Stacks Transactions JS](https://github.com/blockstack/stacks.js/tree/master/packages/transactions) library supports these operations
```
lgalabru, line 103: I think we should point users to @stacks/stacking instead, there's a lot of helpers in there

src/pages/understand-stacks/integrate-stacking.md
```diff
@@ -326,8 +326,10 @@ Stacking delegation requires you to implement a different flow. Until the provid
 
 Below is a description of contract calls required to integrate delegation:
 
-1. Account holders call the [`delegate-stx`](/references/stacking-contract#delegate-stx) method and provide the delegator address, the amount accessible to the delegator, the address rewards should be send to, and the burn height at which the delegation relationship should be terminated
-2. For each delegation relationship that was created, the delegator calls the [`delegator-stack-stx`](/references/stacking-contract#delegate-stack-stx) method to lock up the STX token from the account holder. This method must be called until the delegator locked up enough STX tokens required to participate in Stacking
+1. Account holders call the [`delegate-stx`](/references/stacking-contract#delegate-stx) method and provide the delegator address, the amount accessible to the delegator, the address rewards should be send to, and the burn height at which the delegation relationship should be terminated. This needs to happen only once. The account holder can specify a delegation amount that is higher that their balance
```
lgalabru, line 329: Maybe specify that when the locking is happening, the users should hold the expected amount?

src/pages/understand-stacks/network.md
agraebe: oh, this is a fix that is not related to stacking - it's on the network page. just fixing a broken URL. But yes, we do link to the stacking lib where it makes sense :)

src/pages/understand-stacks/integrate-stacking.md
agraebe: good point

PR #1011, opened 2021-01-04
Mainnet docs and changes

CharlieC3:
## Description

* Adds instructions for running a mainnet node
* Adds instructions for running a mainnet miner
* Separates running a mainnet and testnet miner onto two separate pages
* Updates testnet miner redirect
* Adds new icons for mainnet pages
* Updates stacks-node configuration references

Closes https://github.com/blockstackpbc/devops/issues/545

## Checklist
- [X] [Conventional commits were used](https://www.conventionalcommits.org/en/v1.0.0/)
- [X] Clear code samples were provided

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -609,6 +609,11 @@ async function redirects() {
       permanent: true,
     },
     {
+      source: '/start-mining',
+      destination: '/start-mining/mainnet',
+      permanent: true,
+    },
+    {
       source: '/smart-contracts/overview',
       destination: '/write-smart-contracts/overview',
       permanent: true,
--- a/next.config.js
+++ b/next.config.js
@@ -605,7 +605,7 @@ async function redirects() {
     },
     {
       source: '/mining',
-      destination: '/start-mining',
+      destination: '/start-mining/mainnet',
       permanent: true,
     },
     {
--- a/src/pages/start-mining/mainnet.md
+++ b/src/pages/start-mining/mainnet.md
@@ -1,7 +1,7 @@
 ---
-title: Mine Stacks tokens
-description: Set up and run a miner on the Stacks 2.0 testnet
-icon: TestnetIcon
+title: Mine mainnet Stacks tokens
+description: Set up and run a miner on the Stacks 2.0 mainnet
+icon: MainnetIcon
 experience: beginners
 duration: 10 minutes
 tags:
@@ -13,10 +13,15 @@ images:
 
 ## Introduction
 
-Make sure you've followed the [Running testnet node](/understand-stacks/running-testnet-node) tutorial. Once completed it's only a few more steps to run a proof-of-burn miner on the testnet.
+Make sure you've followed the [Running mainnet node](/understand-stacks/running-mainnet-node) tutorial. Once completed it's only a few more steps to run a proof-of-burn miner on the mainnet.
 
 [@page-reference | inline]
-| /understand-stacks/running-testnet-node
+| /understand-stacks/running-mainnet-node
+
+If you're interested in mining on the testnet, you can find instructions on how to do that here:
+
+[@page-reference | inline]
+| /start-mining/testnet
 
 If you want to learn more about the technical details of mining, please review the mining guide:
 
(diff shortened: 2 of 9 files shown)


agraebe:
vale is making this particular review quite difficult. we could try and increase the annotation "level". I believe it is set to suggestions, warnings, and errors. We could just select "errors"

next.config.js
```diff
@@ -605,7 +605,7 @@ async function redirects() {
     },
     {
       source: '/mining',
-      destination: '/start-mining',
+      destination: '/start-mining/mainnet',
```
agraebe, line 608: we also need a new redirect for `/start-mining` to `start-mining/mainnet`

src/pages/start-mining/mainnet.md
```diff
@@ -0,0 +1,367 @@
...
+
+## Running bitcoind locally
+
+To participate as a miner on mainnet, you must have access to a mainnet bitcoin node. One way to accomplish this is to run bitcoind locally. [Ensure your computer meets the minimum hardware requirements before continuing.](https://bitcoin.org/en/bitcoin-core/features/requirements)
+
+First, download the bitcoind software for your platform from https://bitcoin.org/en/download.
+
+Next, start bitcoind with the following configuration:
+
+```
```
agraebe, line 39: nit: you could use yaml or similar, so the highlighting works

src/pages/start-mining/mainnet.md
```diff
@@ -0,0 +1,367 @@
...
+rpcport=8332
+```
+
+Finally, start bitcoind as follows:
+
+```bash
+bitcoind -conf=path/to/bitcoin.conf
+```
+
+It may take a few days for the node to synchronize with the Bitcoin mainnet.
```
agraebe, line 59: nit: I'm a bit conflicted with the above indication for the timing being 10 minutes. If the reader for the first time looks at this, it'll take days to complete, do I get this right?

src/pages/start-mining/mainnet.md
```diff
@@ -0,0 +1,367 @@
...
+
+## Running a miner
+
+First, a keychain needs to be generated. With this keychain, we'll purchase some BTC from a crytpocurrency exchange, and then use that BTC to start mining.
+
+To get a keychain, the simplest way is to use the `stacks-cli`. We'll use the `make_keychain` command.
```
agraebe, line 65: nit: can link to `/understand-stacks/command-line-interface`

src/pages/start-mining/mainnet.md
```diff
@@ -0,0 +1,367 @@
...
+
+```bash
+npx @stacks/cli make_keychain 2>/dev/null
```
agraebe, line 68: nit: to avoid that the user loses this info, maybe we should store it in a file, like here: https://blockstack-docs-git-mainnet-docs.blockstack.vercel.app/write-smart-contracts/hello-world-tutorial#step-4-create-a-stacks-account

src/pages/start-mining/mainnet.md
```diff
@@ -0,0 +1,367 @@
...
+  }
+}
+```
+
+**Don't lose this information** - we'll need to use the `privateKey` field later on.
+
+The above BTC address will then need to be imported into the BTC network.
+
+```bash
+bitcoin-cli -rpcport=8332 -rpcuser=your-user -rpcpassword=your-password importaddress <btcAddress from JSON above>
```
agraebe, line 90: is this CLI going to be installed with bitcoind?

agraebe (approved):
left some nits but I'd prefer for us to merge this rather quickly, so approving for you to make the calls on what to update or not.

thanks 🙏🏼

src/pages/start-mining/mainnet.md
CharlieC3: Yes that's correct. There may be faster ways we can look into for recommending to users, but it's too close to make a docs change for that now. I'd rather leave it as is, since this is the path we took.

src/pages/start-mining/mainnet.md
CharlieC3: This referenced section seems to prioritize examples using the testnet. While it does explain how to switch between mainnet and testnet usage, I think this redirection might do more harm than good until a clearer distinction is made here. I'll leave this alone for now, since the instructions provided here are correct.

src/pages/start-mining/mainnet.md
CharlieC3: Yup, it's included.

PR #1013, opened 2021-01-05
Add link to running bitcoind

friedger:
## Description
There are a lot of questions about configuring bitcoind from stacks miners. This PR adds a link the more reading and explains the motiviation for the given bitcoind config file.

@agraebe

### DIFF
--- a/src/pages/start-mining.md
+++ b/src/pages/start-mining.md
@@ -47,6 +47,7 @@ bind=0.0.0.0:18333
 rpcbind=0.0.0.0:18332
 rpcport=18332
 ```
+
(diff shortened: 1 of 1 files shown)


CharlieC3 (changes requested):
Good information, just needs to be added to both the testnet and mainnet mining pages and should be good to go!

agraebe:
> Good information, just needs to be added to both the testnet and mainnet mining pages and should be good to go!

@friedger could you add that and we'll merge?

agraebe:
small change, merging conflict, and no response on this for a while. I'm closing but please re-open and update if you have time @friedger

PR #1049, opened 2021-01-20
fix: docs for specifying network when transaction signing

hstove:
A few small changes related to transaction signing. Lots more to improve here but it's better than nothing!

- Update some links that pointed to https://blockstack.org/wallet to https://hiro.so/wallet/install-web.
- Add a notice in the installation section that you have to use a Connect beta (hopefully resolving that soon, just don't want devs to go through the wrong flow)
- Add docs about specifying the `network` option through Connect
- Add docs for getting the current user's STX address
- Updated `onFinish` docs - it now includes a `stacksTransaction` property.

### DIFF
--- a/src/pages/build-apps/guides/transaction-signing.md
+++ b/src/pages/build-apps/guides/transaction-signing.md
@@ -77,7 +77,8 @@ openSTXTransfer({
     name: 'My App',
     icon: window.location.origin + '/my-app-logo.svg',
   },
-  finished: data => {
+  onFinish: data => {
+    console.log('Stacks Transaction:', data.stacksTransaction);
     console.log('Transaction ID:', data.txId);
     console.log('Raw transaction:', data.txRaw);
   },
@@ -97,7 +98,7 @@ interface STXTransferOptions {
     name: string;
     icon: string;
   };
-  finished: (data: FinishedTxData) => void;
+  onFinish: (data: FinishedTxData) => void;
 }
 ```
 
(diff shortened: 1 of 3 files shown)


src/pages/build-apps/guides/transaction-signing.md
```diff
@@ -27,38 +27,58 @@ See the public registry tutorial for a concrete example of this functionality in
...
+
 ## Prompt to transfer STX
 
 Call the `openSTXTransfer` function provided by the `connect` package to trigger the display of a transaction signing prompt for transferring STX:
 
 ```tsx
 import { openSTXTransfer } from '@stacks/connect';
+import { StacksTestnet } from '@stacks/network';
+
+const network = new StacksTestnet();
```
agraebe, line 68: would it make sense to leave a comment suggesting how to change this up for mainnet?

src/pages/build-apps/guides/transaction-signing.md
```diff
@@ -210,24 +236,60 @@ interface ContractCallOptions {
...
+| onFinish        | function         | true     | Callback executed by app when transaction has been signed and broadcasted. [Read more](#onFinish-option)                                                                                                     |     |
 | authOrigin      | string           | false    | URL of authenticator to use for prompting signature and broadcast. Defaults to `https://wallet.hiro.so` for the Stacks Wallet, which is handled by the Stacks Wallet browser extension if installed.         |
+| network         | StacksNetwork    | false    | Specify the network that this transaction should be completed on. [Read more](#network-option)                                                                                                               |
+
+## Getting the signed transaction back after completion {#onFinish-option}
+
+Each transaction signing method from `@stacks/connect` allows you to specify an `onFinish` callback. This callback will be triggered after the user has successfully broadcasted their transaction. The transaction will be broadcasted, but it will be pending until it has been mined on the Stacks blockchain.
```
agraebe, line 262: maybe might be nice to drop a link to the explorer in here?

src/pages/build-apps/guides/transaction-signing.md
```diff
@@ -264,4 +326,4 @@ const MyComponent = () => {
 
 ## Request testnet STX from faucet
 
-You may find it useful to request testnet STX from [the faucet](https://testnet.blockstack.org/faucet) while developing your app with the Stacks testnet.
+You may find it useful to request testnet STX from [the Stacks Explorer sandbox](https://explorer.stacks.co/sandbox?chain=testnet) while developing your app with the Stacks testnet.
```
agraebe, line 329: should we attach the `?chain=` param in links to not?

agraebe (approved):
left some nits, but otherwise looks great. thanks 🙏🏼

src/pages/build-apps/guides/transaction-signing.md
hstove: I added a section at the bottom and links to it in each reference table, but if you'd prefer to be more explicit we can!

src/pages/build-apps/guides/transaction-signing.md
hstove: Yep that's a good idea. Even just sample code for how to link to the explorer.

src/pages/build-apps/guides/transaction-signing.md
hstove: I think it makes sense for testnet. I know the sandbox is testnet only at this time, but if we have a mainnet sandbox (without a faucet obviously) then this would probably be the best URL.

src/pages/build-apps/guides/transaction-signing.md
agraebe: yeah, I'd err on the side of things that could go wrong, will. can already see how devs wouldn't read to the end of the page and be puzzled.. but that's an optimization for sure. your call

PR #1051, opened 2021-01-20
fix: vale.ini location

hstove:
The `vale-action` was failing due to some misconfiguration. We don't need to have a remotely-hosted `.vale.ini` file - that's only needed if the file is not available in your repo. The `.vale.ini` file also had an incorrect `StylesPath`, which caused the pr to fail.


agraebe (commented):
nice! thanks for that!

agraebe (approved):

PR #1057, opened 2021-01-22
Clarify where the output of a tx appears

friedger:
This PR
* changes the last part of the hello world tutorial to clarify where to look for the tx result.

The user should be directed to the explorer to lookup the tx id


agraebe (approved):
lgtm 🙏🏼

PR #1061, opened 2021-01-27
fix: Wrap await in async function

dominicwong617:
## Description

Wrap await in async function in "Step 3: Reviewing account info"

1. Motivation for change?
2. What was changed?
3. Link to relevant issues?

## Checklist

### DIFF
--- a/src/pages/understand-stacks/managing-accounts.md
+++ b/src/pages/understand-stacks/managing-accounts.md
@@ -75,9 +75,9 @@ const stacksAddress = getAddressFromPrivateKey(
   TransactionVersion.Testnet // remove for Mainnet addresses
 );
 
-async function getAccountInfo() {
-  const accounts = new AccountsApi(apiConfig);
+const accounts = new AccountsApi(apiConfig);
 
+async function getAccountInfo() {
   const accountInfo = await accounts.getAccountInfo({
     principal: stacksAddress
   });
@@ -106,7 +106,7 @@ The `balance` property represents the Stacks token balance, as hex-encoded strin
 Proofs, provided as hex-encoded strings, can be removed from the responses by setting the `proof` parameter:
 
 ```js
-async function getAccountInfo() {
+async function getAccountInfoWithoutProof() {
   const accountInfo = await accounts.getAccountInfo({
     principal: stacksAddress,
     proof: 0,
--- a/src/pages/understand-stacks/managing-accounts.md
+++ b/src/pages/understand-stacks/managing-accounts.md
@@ -127,9 +127,9 @@ async function runFaucetStx() {
   const faucetTx = await faucets.runFaucetStx({
     address: stacksAddress,
   });
-};  
-
-console.log(faucetTx);
+  
+  return faucetTx;
+};
 ```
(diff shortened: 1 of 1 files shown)


agraebe:
thanks for submitting this PR. This is an interesting one because we'd need to agree on a convention. If we do this here, we should do this in all other places - which reasons me to believe we shouldn't introduce this.

A developer should be able to put this into a method and run it in a way that makes the most sense. 

Happy to hear what y'all thinking

dominicwong617:
When I did the tutorial, I didn't imagine the code sample to not work as my focus was to learn your libraries. It would be better if developers can focus on learning BlockStack instead of fixing JavaScript code (It might also not be obvious for non-JS developers). I am happy to change the other occurrences as well if needed.

agraebe:
I get the point. I think non-JS devs would be challenged in a variety of places and we'd need to find the right "baseline" for expectations. Without going into all that, a quick fix might be to just wrap the code pieces into methods that could be copy&pasted?

For instance:
```js
async function getAccountInfo() {
  const accounts = new AccountsApi(apiConfig);

  const accountInfo = await accounts.getAccountInfo({
    principal: stacksAddress
  });

  return accountInfo;
}
```

wdyt? can we update the PR to do that?

dominicwong617:
Updated!

src/pages/understand-stacks/managing-accounts.md
```diff
@@ -75,15 +75,15 @@ const stacksAddress = getAddressFromPrivateKey(
   TransactionVersion.Testnet // remove for Mainnet addresses
 );
 
-...
-
+async function getAccountInfo() {
   const accounts = new AccountsApi(apiConfig);
```
agraebe, line 79: I think this was is used in other methods, I'd suggest initializing it outside of this method. so the other examples could work with copy&paste

src/pages/understand-stacks/managing-accounts.md
```diff
@@ -106,24 +106,30 @@ The `balance` property represents the Stacks token balance, as hex-encoded strin
 Proofs, provided as hex-encoded strings, can be removed from the responses by setting the `proof` parameter:
 
 ```js
-const accountInfo = await accounts.getAccountInfo({
-  principal: stacksAddress,
-  proof: 0,
-});
+async function getAccountInfo() {
```
agraebe, line 109: I would rename this method so it is not the same as the one above, eg `getAccountInfoWithoutProof`

agraebe (commented):
I think this would be a good change - left 2 comments. thanks 🙏🏼

src/pages/understand-stacks/managing-accounts.md
dominicwong617: Fixed

src/pages/understand-stacks/managing-accounts.md
dominicwong617: Fixed

dominicwong617:
Updated!

agraebe (approved):
Thanks for the contribution! 👍

PR #1063, opened 2021-01-29
fix: update guidance to use 'canary' npm tag for tx signing

hstove:
Mainly just a better named NPM tag, which we have talked about. Now it's ready.


agraebe (approved):
Lgtm

PR #1064, opened 2021-02-01
docs: add mempool info

agraebe:
Fixes blockstack/docs#1044


src/pages/understand-stacks/technical-specs.md
```diff
@@ -62,6 +62,7 @@ description: Summary of technical specifications of Stacks 2.0
   3. _sending account_ is the account that identifies who is currently executing the transaction: this can change as a transaction executes via the `as-contract` Clarity function
 - Two types of authorizations: standard authorization is where originating account is the same as paying account. _Sponsored_ authorization is where originating account and paying account are distinct. For instance, developers or service providers could pay for users to call their smart-contracts.
 - For sponsored authorization, first a user signs with the originating account and then a sponsor signs with the paying account.
+- Mempool limit for concurrent pending transactions is 25 per account
```
diwakergupta, line 65: @agraebe we should clarify that this limit is _per address_. Without qualification one might interpret this is a global limit.

agraebe: are you alluding to the difference between account and address? we're using them synonymously: https://docs.blockstack.org/understand-stacks/accounts.

Can update if you feel strongly one way or another.

PR #1069, opened 2021-02-07
fix: update refs with bns

agraebe:
Fix mentioned in https://github.com/blockstack/docs/pull/1067#issuecomment-774782960


agraebe:
fyi - I'm merging this to ensure main is building successfully again

PR #1086, opened 2021-02-18
fix: link to client reference

friedger:
This PR
* uses the following link to the client library: https://blockstack.github.io/stacks-blockchain-api/index.html 

@agraebe


agraebe:
this is not the correct link. it should link to: https://blockstack.github.io/stacks-blockchain-api/client/index.html. I wouldn't make this change - the API docs do not link to the client docs, so the reader would be stuck.

I'm assuming this PR was created when the client docs were not committed (which happened sometimes in the past), so the link failed. 

cc @zone117x fyi

PR #1098, opened 2021-03-07
Fix the arguments to call_read_only_contract_func

amadeann:
I was going through the "Hello World" tutorial, and when calling:

`stx call_read_only_contract_func -t <my address> hello-world echo-number 5000 1 <my private key>`

I was getting the error:

`Error: Error calling read-only function. Response 400: Bad Request. Attempted to fetch http://testnet-master.blockstack.org:20443/v2/contracts/call-read/ST16YF3PEV0YE6KER7JSJVSEYY2A3CGF43MZBX44K/hello-world/echo-number and failed with the message: ""`

Called `stx call_read_only_contract_func --help`, and realized the arguments in the tutorial didn't match the expected ones.


agraebe:
thanks for creating this fix. I just merged this fix in a different PR, so closing it out.

🙏🏼

chancelorb:
It looks like this issue is back in the tutorial.

PR #1099, opened 2021-03-08
fix: remove canary info for installing connect for tx signing

hstove:
Now that we've finally closed the Canary PR / branch on Stacks Wallet for Web, we can remove the disclaimer to install the `canary` tag from NPM.


agraebe (approved):
🙏🏼

PR #1101, opened 2021-03-10
Fix get-counter function type in the counter tutorial

amadeann:
`get-counter` function was defined as `public`, which didn't make sense. It should be `read-only`.


agraebe (approved):
thanks!

PR #1116, opened 2021-03-21
Issue 1115.

iamgenchi:
## Description

See [Network Selection Issue](https://github.com/blockstack/docs/issues/1115).

- [x] Fixed flag which was incorrectly set to -H instead of -I


agraebe (approved):
thanks 🙏🏼

PR #1125, opened 2021-03-31
Update integrate-stacking-delegation.md

friedger:
## Description

This PR 
* clarifies that delegated stacking and tasks by tool admins happens usually in two client apps
* add more information about who to run a pool

### DIFF
--- a/src/pages/understand-stacks/integrate-stacking-delegation.md
+++ b/src/pages/understand-stacks/integrate-stacking-delegation.md
@@ -105,7 +105,6 @@ This method calls the [`delegate-stx`](/references/stacking-contract#delegate-st
 
 **Congratulations!** With the completion of this step, you successfully learnt how to use the Stacking library to delegate STX tokens as an account holder.
 
-
 ## Optional: Revoke delegation rights
 
 Delegators will be able to Stack STX tokens on the account holder's behalf until either the set burn height is reached or the account holder revokes the rights.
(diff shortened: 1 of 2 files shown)


src/pages/understand-stacks/integrate-stacking-delegation.md
```diff
@@ -95,13 +99,37 @@ const delegetateResponse = await client.delegateStx({
 // }
 ```
 
-This method calls the [`delegate-stx`](/references/stacking-contract#delegate-stx) method of the Stacking contract.
+This method calls the [`delegate-stx`](/references/stacking-contract#delegate-stx) method of the Stacking contract. Note, that the amount can be higher or lower than the current account balance. Delegation does not yet lock the STX tokens, users can still transfer them.
 
 -> To avoid handling private keys, it is recommended to use the [Stacks Wallet](https://www.hiro.so/wallet) to sign the delegation transaction
 
-## Step 3: Stack STX tokens
+**Congratulations!** With the completion of this step, you successfully learnt how to use the Stacking library to ...
```
agraebe, line 106: could we just make this one sentence? One bullet point seems off because it's not really a list :)

agraebe (approved):
@friedger awesome PR, thanks! I left a miner nit - would be great if you could fix that. feel free to merge right after 🙏🏼

friedger:
Thank you for the review

PR #1144, opened 2021-04-29
feat: microblock docs

agraebe:
This PR adds details related to microblocks to the documentation. These details are included in several pages to ensure discoverability. The PR also includes API details that are in development (with a hint that things could change).

It would be great to release an initial version of macroblocks docs. Let's aim to reduce the scope to that and follow up with more PRs to improve/update.


diwakergupta (changes requested):
High-level comment is to see if we can communicate the ideas even more simply. Here are 2 approaches, I'm sure there are others:

- Problem: Stacks consensus is driven by BTC block production (~every 10min). Is it possible to continue building up the chain _in between_ these leader elections?
  * microblocks are one way to solve this. Leader of current block can build a stream of microblocks, and leader of the next block is incentivized by protocol to confirm them.
- Problem: you can think of transaction finality as a spectrum -- with every confirmed block, the likelihood a transaction is reversed decreases exponentially. Normal transactions offer 1-block finality in ~10min. Is there a lower latency / lower finality option?
  * Microblocks solve this by offering "sub 1-block" finality (a microblock can be confirmed in the current leader's stream in ~10s). In theory these transactions could see more frequent re-orgs; in practice, we don't see much difference (would be nice if we can back this up with some data)

src/pages/understand-stacks/mining.md
```diff
@@ -62,3 +62,15 @@ To calculate the amount of BTC to send miners should:
 
 - Guess the price BTC/STX for the next day (100 blocks later)
 - Guess the total amount of BTCs committed by all miners
+
+## Microblocks
+
+Unlike most existing blockchains, Stacks blocks are not produced atomically. Instead, leaders may dynamically package transactions into a sequence of microblocks as they are received from users.
```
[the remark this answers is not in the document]
agraebe: This explanation is borrowed from the SIP. In this section, the idea was to explain what macroblocks are. Maybe it would be better to frame this as much:
- Anchor blocks "package" transactions and anchor them in a bitcoin block when mined
- Microblocks "stream" transactions to the miner for consideration for new anchor blocks

src/pages/understand-stacks/mining.md
```diff
@@ -62,3 +62,15 @@ To calculate the amount of BTC to send miners should:
...
+
+A leader can "stream" a block over the course of its tenure by selecting transactions from the mempool as they arrive and packaging them into microblocks. These microblocks contain small batches of transactions, which are organized into a hash chain to encode the order in which they were processed. If a leader produces microblocks, then the new chain tip the next leader builds off of will be the last microblock the new leader has seen.
```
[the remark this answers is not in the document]
agraebe: We can simplify I think. The page itself is pretty in-depth, assuming the audience are miners who are interested in the details: https://docs.stacks.co/understand-stacks/mining.

also, @pgray-hiro - this seems to be one of the pages we might consider non-developer-oriented

src/pages/understand-stacks/stacks-blockchain-api.md
```diff
@@ -290,6 +298,65 @@ This API supports [v1.4.6 of the Rosetta specification](https://www.rosetta-api.
...
+      "parent_block_height": 14460,
+      "parent_block_hash": "0xe0d1e8d216a77526ae2ce40294fc77038798a179a6532bb8980d3c2183f58de6",
+      "block_hash": "0x17ceb3da5f36aab351d6b14f5aa77f85bb6b800b954b2f24c564579f80116d99",
+      "txs": ["0x0622e096dec7e2f6e8f7d95f732e04d238b7381aea8d0aecffae026c53e73e05"]
+    }
+  ]
+}
+```
+
+Microblocks are a trdeoff between finality and latency. The initial confirmation of a transaction inside a microblock does not imply finality. Microblocks are subject to reorganization. Thus, the order of processing for confirmed transactions can change. Only once the microblock was added to an anchor block, finality can be assumed.
```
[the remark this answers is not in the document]
agraebe: irt finality: Good point, I can remove this paragraph. We have something similar in a different place. I was over-indexing on the tradeoff details here I'd say.

As for the meta-comment, would love to hear what @pgray-hiro thinks about this. IMO, this page was written in a guide format (not a pure reference) and could include best practices or tradeoff details like that. Ultimately, Patrick should decide on the scoping and format in the future, so I'll let him weigh in :)

agraebe:
> High-level comment is to see if we can communicate the ideas even more simply. Here are 2 approaches, I'm sure there are others:
> 
> * Problem: Stacks consensus is driven by BTC block production (~every 10min). Is it possible to continue building up the chain _in between_ these leader elections?
>   
>   * microblocks are one way to solve this. Leader of current block can build a stream of microblocks, and leader of the next block is incentivized by protocol to confirm them.
> * Problem: you can think of transaction finality as a spectrum -- with every confirmed block, the likelihood a transaction is reversed decreases exponentially. Normal transactions offer 1-block finality in ~10min. Is there a lower latency / lower finality option?
>   
>   * Microblocks solve this by offering "sub 1-block" finality (a microblock can be confirmed in the current leader's stream in ~10s). In theory these transactions could see more frequent re-orgs; in practice, we don't see much difference (would be nice if we can back this up with some data)

I just chatted with Patrick and we'll add a macroblocks overview page. This page will have an introduction with the problem statement, like you proposed. I personally like the first approach you suggested, so I'll add that in my next iteration. Thanks!

zone117x:
@diwakergupta 
> Problem: you can think of transaction finality as a spectrum -- with every confirmed block, the likelihood a transaction is reversed decreases exponentially. Normal transactions offer 1-block finality in ~10min. Is there a lower latency / lower finality option?
Microblocks solve this by offering "sub 1-block" finality (a microblock can be confirmed in the current leader's stream in ~10s). In theory these transactions could see more frequent re-orgs; in practice, we don't see much difference (would be nice if we can back this up with some data)

A few things come to mind. I don't think a direct comparison can be made between microblock sequence numbers "confirmations" and anchor block confirmations. For example, a microblock with sequence number 20 is not at all comparable to an anchor block with 20 confirmations. In fact IIUC, a microblock at _any_ sequence number broadcasted at a given chain tip doesn't really have a higher probability of becoming canonical than the _first_ microblock in a stream. The numbers I had last week showed a ~15% micro-orphan rate, and I don't think that has significant correlation with the microblock sequence number. 

Regarding what we have In practice -- I don't think the STX network has sophisticated miner software, or really anything much more than MVP shipped with the core node. If/when STX fee market and mining matures, I think whatever orphan rates we see now would change. E.g. manipulative RBF transactions that incentivize smart miners to play into certain manipulative schemes with naive users who "trust" micro-confirmations too much. 

I'd like to hear @kantai's and @jcnelson's take.

diwakergupta:
> I don't think a direct comparison can be made between microblock sequence numbers "confirmations" and anchor block confirmations

absolutely, don't think I implied that but lest there was any confusion, I agree we should NOT insinuate that sequence numbers are analogous to block confirmations. All I was trying to saw is that finality for microblock transactions can be thought of as between 0-block and 1-block confirmations (if we think of finality as a spectrum) -- not that the relative position is correlated with the sequence number.

src/pages/understand-stacks/microblocks.md
```diff
@@ -0,0 +1,97 @@
...
+## Transaction states
+
+The Stacks block production model is described in [SIP-001][]. The standard outlines the mechanism by which elected
+block leaders can produce blocks on the Stacks blockchain either by batching transactions or by streaming them.
+Microblocks are the product of the streaming model.
+
+If a block leader has elected to mine microblocks, the leader selects transactions from the mempool and package them
+into microblocks during the current epoch. Transactions included in microblocks don't achieve finality until they're
+included in an anchor block, but can assumed to be highly likely to achieve finality in the order in which the leader
+packaged them.
```
kantai, line 29:
```suggestion
into microblocks during the current epoch. Transactions included in microblocks don't achieve finality until they've
been confirmed by an anchor block, but can be assumed to be highly likely to achieve finality in the order in which the leader
packaged them.
```

src/pages/understand-stacks/microblocks.md
```diff
@@ -0,0 +1,97 @@
...
+While there is a high degree of confidence that streamed blocks are finalized in the order in which a block leader
+selects them, there is still a chance that microblocks can be reorganized in an anchor block. The processing order of
+transactions in microblocks can change depending on on how the microblock is finalized in the anchor block.
+
+-> If a transaction is dependent on a chain state that could by altered by previous transactions with serious
+implications, you should carefully consider whether it should be performed using microblocks.
+
+## Enabling microblocks
+
+Miners can choose to enable or disable microblocks in their mining configuration. As a best practice, you should enable
```
kantai, line 40: I'd replace `you` with `miners`

src/pages/understand-stacks/microblocks.md
```diff
@@ -0,0 +1,97 @@
...
+### Mining
+
+Stacks miners must enable microblocks in their miner configuration to implement the block streaming model. For more
+information, see [mining microblocks][].
+
+## Developing with microblocks
+
+Stacks allows you to choose how any transaction should be included on the blockchain by the miners. This flexibility
+means you can submit transactions that require low latency processing for inclusion in microblocks, and require that
+high stakes transactions wait for anchor block finality.
```
kantai, line 59: I think `highly order-dependent transactions` would be better than `high stakes`.

src/pages/understand-stacks/microblocks.md
```diff
@@ -0,0 +1,97 @@
...
+## Best practices
+
+Working with microblocks is a design decision that you must make for your own application. When working with
+microblocks, the following best practices are recommended.
+
+### Handling nonce
+
+Nonce handling with microblocks is challenging because the next account nonce must take into account any nonce values
+included in microblocks, which may not yet be included in an anchor block. The Stacks Blockchain API
+[provides an endpoint][] to retrieve the next none for a given principal.
```
kantai, line 83: `none` -> `nonce`

kantai (approved):
This looks good to me, just had a couple small comments!

src/pages/understand-stacks/microblocks.md
agraebe: thanks!

src/pages/understand-stacks/microblocks.md
agraebe: 🙏🏼

Issue #1000, opened 2020-12-31
Broken links in ecosystem section


jeffdomke:
To reproduce click on nav item `Stacks token` then click on `Contributing`. Expected: a link. Actual: nothing happens. 

Inspected and there is simply no `<a>` while this page is loaded. It works properly while other pages are loaded so I assume a bug vs. a content change. 

![Screen Shot 2020-12-31 at 9 47 41 AM](https://user-images.githubusercontent.com/767828/103414882-bff4f980-4b4d-11eb-868d-c0c5b5ebd82e.png)

agraebe:
This is interesting. I just tried for myself and it worked. @jeffdomke do you see any errors in the console? Is anything indicating it's loading after you click?


@aulneau is this one of these content blocking issue we saw in the past?

aulneau:
I can reproduce this, i'll assign myself and fix it :)

agraebe:
@aulneau was this fixed?

agraebe:
@aulneau did you have a chance to fix this?

PR #1151, opened 2021-05-11
fix: fixed links in sidebar nav for ecosystem section

pgray-hiro:
## Description

This PR fixes the broken links in the Ecosystem left nav (#1000).  Please review the preview build and ensure that this is the expected look and feel - this change brings the Ecosystem section in line with the other sections so that the links are not always shown in the sidebar.

## Checklist

### DIFF
--- a/src/common/navigation.yaml
+++ b/src/common/navigation.yaml
@@ -130,8 +130,8 @@ sections:
               - path: /language-keywords
               - path: /language-functions
 
-  - title: Ecosystem
-    pages:
-      - path: /overview #pbc et al
-      - path: /stacks-token
-      - path: /contributing
+      - path: /ecosystem
+        pages:
+          - path: /overview #pbc et al
+          - path: /stacks-token
+          - path: /contributing


src/common/navigation.yaml
```diff
@@ -130,8 +130,8 @@ sections:
...
 
-  - title: Ecosystem
-    pages:
-      - path: /overview #pbc et al
-      - path: /stacks-token
-      - path: /contributing
+      - path: /ecosystem
+        pages:
+          - path: /overview #pbc et al
+          - path: /stacks-token
```
agraebe, line 136: I also wonder how relevant content like the token details are 🤔 do we know from analytics how often this page is used? looks like there's no other page cover it outside of the docs (in that detail), so maybe we should keep it for now.

agraebe (approved):

src/common/navigation.yaml
```diff
@@ -130,8 +130,8 @@ sections:
               - path: /language-keywords
               - path: /language-functions
 
-  - title: Ecosystem
-    pages:
-      - path: /overview #pbc et al
-      - path: /stacks-token
-      - path: /contributing
+      - path: /ecosystem
```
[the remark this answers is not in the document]
pgray-hiro: I'd be open to that change but I think it's outside the scope of this fix, I'd like to discuss a larger reorganization of the navigation of the site soon.

src/common/navigation.yaml
pgray-hiro: Fathom seems to have broken in some way in late April but looking at historical analytics the token details is not a hugely viewed page - it's way down the list from pages that contain actual instructional material like the stacking page and tutorials.

PR #1156, opened 2021-05-12
fix: fathom analytics

pgray-hiro:
## Description

Fathom tracking was broken for the site, after talking to Thomas I updated the domain it is pointed at and also established a new dashboard on Fathom that represents the correct domain.

## Checklist

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -901,7 +901,7 @@ module.exports = withFonts(
       trailingSlash: true,
     },
     env: {
-      FATHOM_ID: 'EPNTIXUM',
+      FATHOM_ID: 'FOEMPXUV',
     },
     redirects,
     pageExtensions: ['js', 'ts', 'tsx', 'md', 'mdx'],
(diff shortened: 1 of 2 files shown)


agraebe (approved):
🎉

next.config.js
```diff
@@ -901,7 +901,7 @@ module.exports = withFonts(
       trailingSlash: true,
     },
     env: {
-      FATHOM_ID: 'EPNTIXUM',
+      FATHOM_ID: 'FOEMPXUV',
```
aulneau, line 904: why is this change needed? I'd be worried about losing historic data for the docs

pgray-hiro: We already have a significant gap in the historical analytics due to this breakage.  Since you can't rename a site on Fathom I thought it would be good to have a dashboard with the appropriate `docs.stacks.co` name, no reason we can't keep the old dashboard around for historical reasons though.

aulneau: ah makes sense

aulneau (approved):

PR #1159, opened 2021-05-13
fix: update the Stacks Docs logo per Ginny

pgray-hiro:
## Description

According to @eugeniadigon the font for the logo we are using on the documentation site is wrong, she provided a correct SVG that builds the Stacks logo with the Docs substring.

## Checklist


aulneau (approved):

eugeniadigon:
The logo needs to be slightly smaller, keeping the same height as the current logo

Current logo is 18.281 height
Updated logo is 19.938 height

Thanks

pgray-hiro:
@eugeniadigon I was trying to bring this logo size in line with stacks.co which has a 20px height - WDYT?

eugeniadigon:
> @eugeniadigon I was trying to bring this logo size in line with stacks.co which has a 20px height - WDYT?

Good thinking, but in this case it's not in the same position, so doesn't really look in line with stacks.co, so please do 18.281 height
Thanks!

aulneau:
Decimal units on the web are not really supported well, I'd suggest 18px

pgray-hiro:
Adjusting the width to `156px` gets us a height of `18.3px` - close enough?

eugeniadigon:
> Decimal units on the web are not really supported well, I'd suggest 18px

I understand, I just want to keep it same as what it is right now on the site, which is: 

Screenshot 2021-05-14 at 15 27 22

PR #1254, opened 2021-07-29
docs: Update microblocks doc based on internal discussions

### LINKED ISSUE #1232: Provide guidance on microblocks UX for app integration
Currently we have a bit of stub guidance for how apps should integrate microblocks into their UX here:

https://docs.stacks.co/understand-stacks/microblocks#application-design

> The state of microblock transactions should be carefully communicated to users. No transaction is final until it's included in an anchor block, and your application design should reflect this.

I wonder whether we could start providing more specific guidance here that would be actionable for developers such as @yknl who was asking about this for Xverse.

I also notice that we're suggesting the UX should _not_ reflect finality of transactions until inclusion in anchor blocks whereas @jasperjansz has expressed the opposite approach may be best (that apps _can_ consider transactions final upon inclusion in microblocks and before anchor blocks). So we may need to decide more clearly just what approach we want to recommend.

We're also planning to release our own UX implementations in the wallets and explorer soon, which consider transactions as rather final upon inclusion in microblocks – but also distinguish between the two states with visual treatment in the explorer only. Perhaps we could simply lay out our two distinct approaches for the wallets and explorer, explaining our rationale for each.

See https://www.figma.com/proto/i2H9felRYWAZQgsfX7EpZz/microblocks?node-id=39%3A4125

cc @pgray-hiro @jasperjansz

pgray-hiro:
## Description

This PR updates the microblocks documentation with information from Aaron in the microblocks discussion in #1232. This is an initial attempt to provide a set of best-practices for how to display pending state represented by microblocks in different kinds of applications.

Closes #1232.

## Checklist

### DIFF
--- a/src/pages/understand-stacks/microblocks.md
+++ b/src/pages/understand-stacks/microblocks.md
@@ -12,7 +12,7 @@ images:
 Microblocks are a protocol-level feature of the Stacks blockchain that solve the technical challenge of transaction
 latency. Because each Stacks block is anchored to a Bitcoin block through the [Proof-of-Transfer consensus mechanism][],
 Stacks is necessarily limited to the same block times as the Bitcoin network. Microblocks allow the Stacks blockchain to
-perform state transitions with a high degree of confidence between final settlements against Bitcoin.
+perform state transitions between anchor blocks.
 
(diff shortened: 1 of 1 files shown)


agraebe (approved):
awesome!

PR #1270, opened 2021-10-03
Fix path to config file

friedger:
## Description
Following the steps of the documentation to run a mainnet node results in `Error: No such container: stacks-blockchain`

The problem is that the docker image crashes without notifying the user due to `-d` option. The crash is caused by a wrong path to the config file.

This PR fixes the path.

Thank you to @d1gitalflow for finding the problem and notifying.

## Checklist

- [x] People were tagged for review @agraebe


agraebe (approved):
thanks for the fix!

PR #1286, opened 2022-01-14
removed duplicate file

criadoperez:
Removed duplicate file index.md:

https://github.com/blockstack/docs/blob/master/src/pages/index/index.md
https://github.com/blockstack/docs/blob/master/src/pages/index.md

This file is currently being used in the link "Edit this page on Github" on this page: https://docs.stacks.co/ , but will instead use the newer version 
https://github.com/blockstack/docs/blob/master/src/pages/index/index.md as soon as the PR https://github.com/blockstack/docs/pull/1285 is merged.


jhammond2012 (approved):
🚀

criadoperez:
@jhammond2012 something is not right here... It seems the menu at the left of the screen is still using the old file, so it doesn't work. Also https://docs.stacks.co/ is loading the deleted file instead of the new one. Any idea on how to change it to the new one?
We could revert this commit but I think it makes more sense to correct the issue instead.

PR #1287, opened 2022-01-15
stacks node with own btc node

criadoperez:
Added optional documentation on how to run a stacks node using your own bitcoin node.


friedger (approved):
Looks good!

PR #1290, opened 2022-01-20
fix: moving home page file

criadoperez:
Moving the home page from "/" to "/index/"


criadoperez:
@jhammond2012 another way to solve this would be to eliminate the duplicate in `docs/src/pages/index/index.md`and leaving `docs/src/pages/index.md` as it is. Is this compatible with the new support for transtralation? If it is it's a simpler solution...

jhammond2012:
@criadoperez I opened a PR [here](https://github.com/stacks-network/docs/pull/1292). What do you think about this solution?

PR #1297, opened 2022-02-28
Fix/criadoperez3

criadoperez:
Updated obsolete repositories


criadoperez (commented):
@jhammond2012 
The correct value should be:
https://github.com/stacks-network/gaia
instead of 
https://github.com/blockstack/gaia
however  "prettier" fails when it's changed to stacks, so I'm leaving the original text for the moment.

PR #1300, opened 2022-03-09
sub folder revert, fix for URL index 404

jhammond2012:
## Description
This PR is reverting the sub-folder setup that was completed for localization. I have moved all pages in the previous arrangement inside the /en folder. Each language will have its own folder. (es, fr, etc.)


criadoperez (approved):
Looking good. Thanks.

PR #1333, opened 2022-04-01
Fix grammar in proof-of-transfer.md

lukechilds:
"Additionally, By" => "Additionally, by"

PR #1335, opened 2022-04-01
Fix/action stylecheck

criadoperez:
## Description

Github action stylecheck was only running when `.md` files where changed, but checks themselves were run also on `.ts` files also.

This was producing undesired effects like on this PR](https://github.com/stacks-network/docs/pull/1333) that only modified a `.md` file, but tests failed because of a change on a `.ts`file done on a previous commit.

To fix this and avoid future fails, the stylecheck action will now run whenever `.md` or `.ts` files are modified.


jhammond2012 (approved):
🚀

PR #1357, opened 2022-04-22
Feat/faqs

criadoperez:
Added first FAQs, starting with why is my tx pending and what is nonce.


criadoperez:
Content of the first FAQs is here, but I suggest we add a layout that is more appropriate for an FAQ section. Something like [this](https://nextjs.org/docs/faq).

jhammond2012 (approved):

Hero-Gamer:
Nice!

PR #1387, opened 2022-05-16
added 'id' in next.config.js

criadoperez:


criadoperez:
Adding `id` language in next.config.js breaks the translation and displays the english version instead.

criadoperez:
Verified that going to https://docs.stacks.co redirects correctly to the correct locale when its set in the browser. 
So for example, viewing for a browser with spanish locale will redirect to https://docs.stacks.co/es

criadoperez:
Closing as migration to Discosaurus makes this issue irrelevant .

Issue #1458, opened 2022-06-15
category json files need to be included in Crowdin translations


criadoperez:
**Describe the issue**

Files `_category_.json` needs to be included in crowdin for translation.

Currently only the markdown files are.

PR #1459, opened 2022-06-15
Activate ES language

criadoperez:


criadoperez:
Watiting for #1458 to be resolved before we activate another language.

PR #1542, opened 2022-09-05
Description of divide and more examples

c0d3rk1d:
## Description

I am learning clarity; attending clarity bootcamp 4 now. So, last a week or so I am on this document everyday.
I came across a few errors on the document, and I would like to add more examples, while I am learning.

While experimenting with **/ (divide)**, I realized that it is not an "integer division". According to wolfram and wikipedia, in integer division, the floor function is applied to the result.

https://mathworld.wolfram.com/IntegerDivision.html
https://en.wikipedia.org/wiki/Division_(mathematics)#Of_integers

Clarity doesn't apply the floor function, it simply returns the integer part of the division, So I changed the description of the **/ (divide)** accordingly.

And I added more examples to 4 Arithmetic Operations.

### DIFF
--- a/docs/write-smart-contracts/clarity-language/language-functions.md
+++ b/docs/write-smart-contracts/clarity-language/language-functions.md
@@ -5,7 +5,7 @@ tags:
   - clarity
 ---
 
-## Functions 
+## Functions
 
 Detailed list of all functions for the Clarity language.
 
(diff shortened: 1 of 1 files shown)


criadoperez:
Thanks for you contributions @c0d3rk1d ! And congratulations for attending the Clarity Bootcamp.

criadoperez (approved):

PR #1549, opened 2022-09-19
log2 description and examples

c0d3rk1d:
## Description

- Log2 fails on 'Zero' 
- There is no rounding, but striping out the decimal point from result
- Indentation of the output in the example, for readability

### DIFF
--- a/docs/write-smart-contracts/clarity-language/language-functions.md
+++ b/docs/write-smart-contracts/clarity-language/language-functions.md
@@ -185,7 +185,7 @@ Returns the largest integer that is less than or equal to the square root of `n`
 #### output: `int | uint`
 #### signature: `(log2 n)`
 #### description:
-Returns the _integer part_ of the power to which the number 2 must be raised to to obtain the value `n`. Fails on zero or a negative numbers.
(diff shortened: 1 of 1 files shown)


c0d3rk1d:
`<1` if check from VM code at 
https://github.com/stacks-network/stacks-blockchain/blob/master/clarity/src/vm/functions/arithmetic.rs#L230

criadoperez (approved):
Tested this and you are right @c0d3rk1d . Thanks for clarifying this!

criadoperez:
> `<1` if check from VM code at https://github.com/stacks-network/stacks-blockchain/blob/master/clarity/src/vm/functions/arithmetic.rs#L230

In my opinion the text `log2 must be passed a positive integer` in the code line you pointed out can be a bit ambiguous and should be changed.

Event though some don't consider 0 to be a positive integer, mathematically speaking it is. So it's unclear.
In my opinion, this should be replaced with something like `log2 must be passed a positive integer equal or larger than 1` or `log2 must be passed a positive integer larger than 0`.

I recommend you open a PR on that repository, pointing out this issue you detected. It will be interesting to see how the rest see it.

PR #1551, opened 2022-09-30
Update software.md

BalogunMalikO:
## Description

There was a typographical error in the documentation, "sintax" was used instead of "syntax"


BalogunMalikO:
hello admin kindly look into the change I'm trying to make

PR #1554, opened 2022-10-08
Add uncheck error information for secp256k1-verify

friedger:
## Description

This PR adds information that the sec256k1-verify function throws an unchecked error if the input arguments have a wrong buffer length.


obycode (approved):
LGTM

PR #1563, opened 2023-01-08
Remove neptune

friedger:
This PR
* removes Neptune from wallets section because it is not updated and does not provide good UX for new users looking for a wallet


obycode (approved):
lgtm!

Issue #1575, opened 2023-02-17
Add Clarity 2 functions and keywords references


hugocaillard:
[clarity/language-functions](https://docs.stacks.co/docs/clarity/language-functions) - [clarity/language-keywords](https://docs.stacks.co/docs/clarity/language-keywords)
---

Developers can write Clarity 2 since a few weeks, the doc should reflect that.
These page could include a Clarity version selector

wileyj:
Good idea - would you be willing to PR the change here? not sure how soon we'll be able to address this on our end

hugocaillard:
@wileyj Sure, you can assign it I'm taking care of it today

PR #1576, opened 2023-02-20
Feat: handle clarity 2 generated api references

hugocaillard:
## Description

Fix: #1575 

- [x] Dynamic pages for language-functions and language-keywords (based on the [stacks-blockchain generated docs)](https://github.com/stacks-network/stacks-blockchain/blob/master/.github/actions/docsgen/Dockerfile.docsgen#L13-L14)
- [x] Proper support for Clarity language in Prism
- [x] Removed some dead code

Screenshot 2023-02-21 at 01 06 24

Note: functions and keywords references now display a line: "introduce in: Clarity <version>".
We could have implemented a `select` to display only functions available in a given Clarity version but it's not trivial because docusaurus table of content aren't dynamic.
The other option would have been to have sub-menus for version "Clarity > Functions > Clarity 1 | 2", but it would have an extra click.
After implementing both, I decided to remove the complexity, assuming that >80% of the time, developers will want to see the latest version (and we specify if a function is available in a given version)


src/components/ApiReferences/_helpers.js
```diff
@@ -0,0 +1,6 @@
+const idPattern = /[^A-Za-z0-9\-]/g;
+export const nameToId = (name, version) => {
+  let slug = name.trim().replaceAll(" ", "-").replace(idPattern, "");
+  let suffix = version === "Clarity1" ? "" : `-${version}`.toLowerCase();
```
hugocaillard, line 4: The suffix is displayed in the url, for example:
```
/clarity/language-functions#element-at
```
To avoid broken links, this method generates the same slugs as docusaurus natively does. So with this new PR, all of the links will still work.

Because Clarity 2 introduces aliases (`element-at?` and `index-of`), I included the clarity version in the anchor, except for Clarity 1 methods.
So 
- `element-at` anchor will be `#element-at`
- `element-at?` anchor will be `#element-at-clarity2`

If clarity 3 introduces somethings like `element-at!` it would be `#element-at-clarity3`

docusaurus.config.js
```diff
@@ -2,7 +2,7 @@
 // Note: type annotations allow type checking and IDEs autocompletion
 
 const lightCodeTheme = require("prism-react-renderer/themes/github");
-const darkCodeTheme = require("prism-react-renderer/themes/dracula");
+const darkCodeTheme = require("prism-react-renderer/themes/vsDark");
```
hugocaillard, line 5: Know that we have proper support for Clarity 2, I didn't like the way the dracula theme displays native function (in _italic_). So here's a test for vsDark.
It looks nice on my macbook pro but less nice on an other display I have (background can be a bit greenish). But developers are used to this background.

before/after
Screenshot 2023-02-21 at 00 55 33
Screenshot 2023-02-21 at 00 55 18

Let me know what you think, I'm happy switching back to dracula as well

src/components/ApiReferences/FunctionsReferences.jsx
```diff
@@ -0,0 +1,51 @@
...
+import { ReactMarkdown } from "react-markdown/lib/react-markdown";
+import CodeBlock from "@theme/CodeBlock";
+import MDXContent from "@theme/MDXContent";
+import MDXHeading from "@theme/MDXComponents/Heading";
+
+import { nameToId } from "./_helpers";
+import { functions } from "../../_data/clarity-reference.json";
+
+export const FunctionsReferences = ({ toc }) => {
+  return functions.sort().map((func) => {
```
hugocaillard, line 11: The generated doc sorting doesn't make much sense

src/components/ApiReferences/KeywordsReferences.jsx
```diff
@@ -0,0 +1,45 @@
...
+import { ReactMarkdown } from "react-markdown/lib/react-markdown";
+import CodeBlock from "@theme/CodeBlock";
+import MDXContent from "@theme/MDXContent";
+import MDXHeading from "@theme/MDXComponents/Heading";
+
+import { nameToId } from "./_helpers";
+import { keywords } from "../../_data/clarity-reference.json";
+
+export const KeywordsReferences = ({ toc }) => {
+  return keywords.map((keyword) => {
```
hugocaillard, line 11: I manually sorted the keywords in the json into something coherent

docusaurus.config.js
wileyj: I have no opinion to be honest - it's a cosmetic change so if someone doesn't like it we can address in a different issue. the change here seems fine to me, visually it still looks fine as well.

wileyj:
looks good to me. any comments @kenrogers before i merge?

PR #1590, opened 2023-04-25
Remove outdated instructions

sneveu:
Just ran through the Hello Stacks tutorial and noticed some outdated instructions.

Changes listed below:

- Tailwind v3 uses the `jit` option by default now so the mode option has been removed.
- AuthDetails was being imported from the `@stacks/connect` package but no longer exists as an export (and wasn't being used anyway) so created compilation errors
- Remove duplicated network variable since it's already been asked to be added below the state declarations


wileyj:
this has been replaced by https://github.com/stacks-network/docs/blob/master/guides-and-tutorials/hello-stacks-quickstart-tutorial.md

if you still see a problem, please reopen

PR #1611, opened 2024-03-07
Fix nit around online tooling.

AshtonStephens:
## Description

Change the guidance around encoding tools to use local method:

```bash
python3 -m venv .venv
source .venv/bin/activate
pip3 install base58
sudo bash -c 'cat <<EOF> ./encode.py
from sys import argv
import base58
print(base58.b58encode(argv[1]).decode("utf-8"))
EOF'
python3 ./encode.py PRIVATE_KEY
```


cuevasm (approved):

wileyj (approved):
shipit

PR #1612, opened 2024-03-09
"clarinet integrate" is deprecated. Use 'clarinet devnet start' instead

VGabriel45:
## Description

"clarinet integrate" shows as a deprecated command, could replace it with "clarinet devnet start" which is recommended.


wileyj:
addressed in https://github.com/stacks-network/docs/pull/1629

PR #1621, opened 2024-04-17
Update storage reqs to 350GB and other fixes

wileyj:
In testing, 250GB *is ok* - but on a Debian server based system (i.e. not a lot of extra packages installed), the chainstate + the os takes around 220GB. 

the chainstate itself is around 190GB today. 

updating to 350GB should be considered the new minimum for 6-12 months of chain growth.


wileyj:
actuallly hold for a few - i think i found something else. converting to draft for now

saralab (approved):

PR #1626, opened 2024-05-16
Chore: update bootstrap_nodes config to use current default values

wileyj:
`bootstrap_nodes` config setting in the mining docs was still set to use the old values. 
this updates to use the current default values


diwakergupta (approved):
shipit!

PR #1633, opened 2024-07-24
fix: docs to `generate-stacking-signature`

aldur:
## Description

Trying to follow the docs to generate a new signer signature through the CLI was failing because of mismatched CLI parameters (in the docs and in the executable).

This PR fixes that, by changing changes `topic` to `method` and adds the missing cli arguments for:

--max-amount <MAX_AMOUNT>
--auth-id <AUTH_ID>

To keep things DRY, I have removed the repetition of the parameters description and just referenced the appropriate section.

Lastly, I added the `--json` flag, since it makes it easier to provide the generated signature to lockstacks.

### DIFF
--- a/nakamoto-upgrade/signing-and-stacking/stacking-flow.md
+++ b/nakamoto-upgrade/signing-and-stacking/stacking-flow.md
@@ -725,30 +725,34 @@ The CLI command is:
 
 ```bash
 stacks-signer generate-stacking-signature \
-  --pox-address bc1... \
-  --reward-cycle 100 \
-  --config ./config.toml \
+  --method stack-stx \
+  --max-amount 1000000000000 \
+  --auth-id 195591226970828652622091037492597751808 \
   --period 1 \
-  --topic stack-stx
+  --reward-cycle 100 \
+  --pox-address bc1... \
+  --config ./config.toml \
+  --json
 ```
 
-These arguments are:
+These arguments match those described in section [Overview of signer keys and signatures](#overview-of-signer-keys-and-signatures), with the addition of:
 
-* pox-address: the BTC address where the signer wants to receive Stacking rewards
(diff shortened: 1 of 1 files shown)


nakamoto-upgrade/signing-and-stacking/stacking-flow.md
```diff
@@ -725,30 +725,34 @@ The CLI command is:
 
 ```bash
 stacks-signer generate-stacking-signature \
-  --pox-address bc1... \
+  --method stack-stx \
+  --max-amount 1000000000000 \
+  --auth-id 195591226970828652622091037492597751808 \
```
wileyj, line 730: where does this auth-id come from?

aldur: The source code states that it is:

> A unique identifier to prevent re-using this authorization

The docs:

> a random integer that prevents re-use of a particular signature, similar to how nonces are used with transactions

I generated it through the command also listed there:

```bash
openssl rand -hex 16 | awk '{ print strtonum("0x"$0) }'
```

I am torn on whether the code snippet should provide examples that the executable would accept, instead of placeholders like in the `--help` for the CLI that force the user to change them. Especially for the fields we are sure the user will have to customise and/or that might provide security troubles.

I can change that to a placeholder, like `--auth-id <AUTH_ID>`. What do you think?

aldur: Also, I am considering suggesting generating the `nonce` (if we have to) with this:

```bash
python3 -c 'import secrets; print(secrets.randbits(128))'
```

It relies on Python's standard library only and does not go into edge-cases of whether `gawk` correctly handles big numbers (spoiler, it probably doesn't and we would need at least to add `-M` to its invocation).

wileyj: both of those are valid options, i just think it's important to specify *what* this field needs, either as a comment or explanation before the command example. 
I see your addition below, thanks!

wileyj (approved):

wileyj:
@kenrogers any objection to merging here? changes look good to my eyes

kenrogers:
Looks great to me, good to merge. Thanks for the fix @aldur 🙏🏻

PR #1636, opened 2024-08-22
fix: add `stacker = true` to configs

obycode:
## Description

This `stacker = true` is required for nodes attached to stackers. It was mentioned in one paragraph, but it should be in the sample configs to ensure it is not missed.


kenrogers (approved):
LGTM, thanks Brice 🙏🏻

PR #1645, opened 2024-09-27
Adding undocumented setting

wileyj:
## Description

Learned of a new signer option `block_proposal_timeout_ms` - useful in case a miner is slow to broadcast blocks, this allows the singer to wait longer before rejecting a block proposal from a valid miner. 

this setting is more relevant to distributed testnets/mocknets than mainnet.


BowTiedDevOps (approved):
LGTM

PR #1652, opened 2024-10-09
update testnet configs to match stacks-core repo

wileyj:
some minor chagnes- removed unused fields from the burnchain section and update the 3.0 block height in the sample configs to match https://github.com/stacks-network/stacks-core/pull/4995


BowTiedDevOps (commented):
LGTM

BowTiedDevOps (approved):

PR #1653, opened 2024-10-10
fix: remove unused config options from examples

obycode:
## Description

An upcoming update will error out if there are unknown fields in the config file. This removes some old fields from the examples.


BowTiedDevOps (approved):
LGTM

PR #1655, opened 2024-10-14
Remove Clarity Function Argument

BowTiedWoo:
The inputs of the `stx-transfer?` clarity function had an extra field named `buff`, which is not available for this function.


BowTiedDevOps (approved):
LGTM

PR #1658, opened 2024-10-16
Add Signer Quickstart Page

BowTiedDevOps:
## Description

Add a single page of documentation with condensed information and commands on how to quickly run a signer and become a stacker in order to make it easier to become an active signer.

This PR will also focus on simplifying the information provided on the existent signer/stacker documentation for users that will read the detailed explanations.


guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,548 @@
...
+}
+```
+
+From this file, you'll need the `privateKey` value.
+
+## Step 2 - Set Up Your Stacks Signer
+
+### Download the stacks-signer binary
+
+Official binaries are available from the [Stacks Core releases page on Github](https://github.com/stacks-network/stacks-core/releases). Each release includes pre-built binaries. Download the [latest signer release ZIP file](https://github.com/stacks-network/stacks-core/releases/tag/signer-2.5.0.0.5.2) for your server’s architecture and decompress it. Inside of that folder is a `stacks-signer` binary. Currently, the latest mainnet and testnet versions that you have to use differ.
```
wileyj, line 76: This one is a little tough to be accurate with because:
- the signer may be released standalone from the stacks-node, in which case your links are fine
- but, the signer may be included in the stacks-core release - in which case the links will be wrong. 

i.e. when 3.0 is released - there will not be a `signer-3.0.0.0.0` tag, it will just be `3.0.0.0.0`

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,548 @@
...
+
+Assuming a `Linux x64` machine, the commands to download and uncompress the signer binary look like this:
```
wileyj, line 78: I would be more specific here and mention glibc (if a user has a musl sytem, the links will not work for them). 
counterpoint is that users using musl by default will generally know they are and recognize the link won't work for them. 

i would default to being clear these are for glibc systems

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,548 @@
...
+
+```bash
+~/stacks-signer/stacks-signer run -c ~/stacks-signer/signer-config.toml
+```
+
+## Step 3 - Set Up Your Stacks Node
+
+### Download the stacks-node binary
+
+Official binaries are available from the [Stacks Core releases page on Github](https://github.com/stacks-network/stacks-core/releases). Each release includes pre-built binaries. Download the [latest node release ZIP file](https://github.com/stacks-network/stacks-core/releases/latest) for your server’s architecture and decompress it. Inside of that folder is a `stacks-node` binary. Currently, the latest mainnet and testnet versions that you have to use differ.
```
wileyj, line 244: I'm not sure this is needed - once 3.0 is released (soon), the versions will align

guides-and-tutorials/running-a-signer/signer-quickstart.md
wileyj: same as below comment - `Currently, the latest mainnet and testnet versions that you have to use differ.` is not needed since versions will align on 3.0 release

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,548 @@
...
+cd ~/stacks-node
+
+# Download the node binary zip
+wget https://github.com/stacks-network/stacks-core/releases/download/2.5.0.0.7/linux-glibc-x64.zip
+
+# Unzip the node binary archive
+unzip linux-glibc-x64.zip
+
+# Remove unwanted files
+find ~/stacks-node -type f ! -name 'stacks-node' -delete
```
wileyj, line 263: i wouldn't add this, since the only benefit is saving some disk space (marginal)

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,548 @@
...
+cd ~/stacks-signer
+
+# Download the signer binary zip
+wget https://github.com/stacks-network/stacks-core/releases/download/3.0.0.0.0-rc5/linux-glibc-x64.zip
+
+# Unzip the signer binary archive
+unzip linux-glibc-x64.zip
+
+# Remove unwanted files
+find ~/stacks-signer -type f ! -name 'stacks-signer' ! -name 'keychain.json' -delete
```
wileyj, line 113: i wouldn't add this, since the only benefit is saving some disk space (marginal)

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,548 @@
...
+stacker = true
+
+[burnchain]
+chain = "bitcoin"
+mode = "mainnet"
+peer_host = "bitcoind.stacks.co"
+username = "blockstack"
+password = "blockstacksystem"
+rpc_port = 8332
+peer_port = 8333
```
aldur, line 316: I have a node configured this way and it seems to fall behind quite often (please reach out and we can look at the data).

If that's the case: shouldn't we suggest users to run their own Bitcoin nodes?

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,548 @@
...
+{% tabs %}
+
+{% tab title="Mainnet" %}
+
+```bash
+# Set environment variables
+AUTH_TOKEN=<your_token> # Used for signer-node authentication, same token as the one set up in the signer configuration
+
+# Create the node's configuration file
+cat <<EOF> ~/stacks-node/node-config.toml
```
aldur, line 301: Please enable monitoring as well and instruct users on how to observe through Grafana (see other post). 

`prometheus_bind` is in testnet config but not in main net.

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,548 @@
...
+```
+
+The generated JSON can be then copy-pasted directly in the [Lockstacks](https://lockstacks.com) website mentioned in the next step.
+
+## Step 6 - Start stacking
+
+The simplest route is to solo stack. You can do that by using [Lockstacks](https://lockstacks.com). Click on the 'Stack Independently' button and follow the instructions there.
+
+If you would like to learn more about solo stacking or running a pool operator, take a look at the [Stack STX](https://docs.stacks.co/guides-and-tutorials/stack-stx) guide.
+
```
aldur, line 548: Step 7: monitor things, please link to monitoring.

guides-and-tutorials/running-a-signer/signer-quickstart.md
BowTiedDevOps: I could add some comments related to that, but I don't want to complicate this page too much - it's meant to be as simple as possible in order to get users to run a signer without too much hassle. Perhaps we could point them at one of your btc nodes @wileyj?

guides-and-tutorials/running-a-signer/signer-quickstart.md
BowTiedDevOps: @wileyj could you please give a second opinion on this? As stated in a previous reply, my point was to simplify the instructions as much as possible. I'm inclined to skip a monitoring step here as that's a dedicated page under the same section, and I'm thinking it would not align with the scope of this page.

BowTiedDevOps:
I'll wait for the stable 3.0 version to be released, after that I'll commit with the same version across this whole page and remove the unnecessary tabs where that's the case

guides-and-tutorials/running-a-signer/signer-quickstart.md
aldur: That would be an ideal solution, we just need to make sure the default setup is robust.

guides-and-tutorials/running-a-signer/signer-quickstart.md
aldur: Thank you! As we wait for @wileyj let me clarify that I'd just a link to the existing guide, no need to replicate the content here. The goal should be for users to run signers, but do that reliably.

wileyj: > Thank you! As we wait for @wileyj let me clarify that I'd just a link to the existing guide, no need to replicate the content here. The goal should be for users to run signers, but do that reliably.

i think  a link here is warranted, if only as a link and not a required step for the quick-start docs.it's relevant and helps the user strengthen a deployment if they desire. 

particularly since we're already linking users to other things they may do with their signer (i.e. stacking), i'm on team "let's add monitoring link here"

guides-and-tutorials/running-a-signer/signer-quickstart.md
wileyj: debatable if it should be commented or not, or maybe something like this:
```
prometheus_bind = 127.0.0.0.1:30001 ## monitoring host:port
```

guides-and-tutorials/running-a-signer/signer-quickstart.md
wileyj: @aldur @BowTiedDevOps that burnchain section is incredibly outdated. 
if anything, i would use the stacks.org bitcoin node, or better add a section to use a local bitcoin node. 

@aldur to your issue specifically - i've only seen that when geographic latency and endpoint CPU/num conns is an issue. 
ex: for whatever reason, hosts in CET have problems staying connected to btc hosts we have in EST

guides-and-tutorials/running-a-signer/signer-quickstart.md
wileyj: since we're re-aligning the release process for the signer, i'm resolving this

guides-and-tutorials/running-a-signer/signer-quickstart.md
wileyj: the link structure is fine after today - but teh version should be updated to reflect the latest signer tag: `signer=3.0.0.0.0.0`

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,435 @@
...
+node_host = "127.0.0.1:20443"
+endpoint = "127.0.0.1:30000"
+network = "testnet"
+db_path = "$HOME/stacks-signer/data/signer.sqlite"
+auth_password = "$AUTH_TOKEN"
+stacks_private_key = "$PRIVATE_KEY"
+metrics_endpoint = "127.0.0.1:9154"
+block_proposal_timeout_ms = 180000
+EOF
+```
```
wileyj, line 137: since the file with the key is being saved, we can entirely do away with the hard-code env vars. 
ex:
```
$ cat test.json 
{
  "privateKey":"bar"
}
$ cat <<EOF> ./test.toml
node_host = "127.0.0.1:30000"
stacks_private_key = "$(jq -r  '.privateKey' test.json)
auth_password = "$(openssl rand -base64 12)"
EOF
$ cat test.toml
node_host = "127.0.0.1:30000"
stacks_private_key = "bar
auth_password = "8ClPmSsvGQ/w+MBo"
```

wileyj: not sure which would be easier for the target audience, but it's an option

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,435 @@
...
+rpc_bind = "0.0.0.0:20443"
+p2p_bind = "0.0.0.0:20444"
+prometheus_bind = "127.0.0.1:9153" 
+bootstrap_node = "02196f005965cebe6ddc3901b7b1cc1aa7a88f305bb8c5893456b8f9a605923893@seed.mainnet.hiro.so:20444,02539449ad94e6e6392d8c1deb2b4e61f80ae2a18964349bc14336d8b903c46a8c@cet.stacksnodes.org:20444,02ececc8ce79b8adf813f13a0255f8ae58d4357309ba0cedd523d9f1a306fcfb79@sgt.stacksnodes.org:20444,0303144ba518fe7a0fb56a8a7d488f950307a4330f146e1e1458fc63fb33defe96@est.stacksnodes.org:20444"
+stacker = true
+
+[burnchain]
+chain = "bitcoin"
+mode = "mainnet"
+peer_host = "bitcoin.stacks.org"
```
wileyj, line 229:
```suggestion
peer_host = "bitcoin.mainnet.stacks.org"
```

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,435 @@
...
+  --pox-address 19tg... \
+  --config ~/stacks-signer/signer-config.toml \
+  --json
+```
+
+The generated JSON can be then copy-pasted directly in the [Lockstacks](https://lockstacks.com) website mentioned in the next step.
+
+## Step 6 - Start stacking
+
+The simplest route is to solo stack. You can do that by using [Lockstacks](https://lockstacks.com). Click on the 'Stack Independently' button and follow the instructions there.
```
wileyj, line 429: is this the easiest? since we're already downloading the binaries, we'll have `blockstack-cli` that would be easier IMO, at the expense of a little more text to this doc

guides-and-tutorials/running-a-signer/signer-quickstart.md
wileyj: @aldur opinion here?

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,435 @@
...
+bootstrap_node = "02196f005965cebe6ddc3901b7b1cc1aa7a88f305bb8c5893456b8f9a605923893@seed.mainnet.hiro.so:20444,02539449ad94e6e6392d8c1deb2b4e61f80ae2a18964349bc14336d8b903c46a8c@cet.stacksnodes.org:20444,02ececc8ce79b8adf813f13a0255f8ae58d4357309ba0cedd523d9f1a306fcfb79@sgt.stacksnodes.org:20444,0303144ba518fe7a0fb56a8a7d488f950307a4330f146e1e1458fc63fb33defe96@est.stacksnodes.org:20444"
+stacker = true
+
+[burnchain]
+chain = "bitcoin"
+mode = "mainnet"
+peer_host = "bitcoin.stacks.org"
+
+[connection_options]
+auth_token = "$AUTH_TOKEN"
```
wileyj, line 232: note above about using openssl for this auth_token. if that above idea is used, this will need to reflect that change

guides-and-tutorials/running-a-signer/signer-quickstart.md
wileyj: target audience of this section should be the deciding factor i think.

guides-and-tutorials/running-a-signer/signer-quickstart.md
```diff
@@ -0,0 +1,435 @@
...
+```bash
+# Set environment variables
+AUTH_TOKEN=<your_token> # Used for signer-node authentication, same token as the one set up in the signer configuration
+
+# Create the node's configuration file
+cat <<EOF> ~/stacks-node/node-config.toml
+[node]
+working_dir = "$HOME/stacks-node/data"
+rpc_bind = "0.0.0.0:20443"
+p2p_bind = "0.0.0.0:20444"
```
wileyj, line 253: not sure if we want this to be open on all interfaces. i think we should default to private instances, and leave it up to the user (i.e. advanced users)if they want to expose the ports outside of localhost. 
```suggestion
rpc_bind = "127.0.0.1:20443"
p2p_bind = "0.0.0.0:20444"
```

not sure if we want this to be open on all interfaces. i think we should default to private instances, and leave it up to the user (i.e. advanced users)if they want to expose the ports outside of localhost. 

at the very least, for the signer we shojldn't allow external RPC requests

guides-and-tutorials/running-a-signer/signer-quickstart.md
BowTiedDevOps: Might not be the easiest to execute, as it would mean navigating to another website and doing the transaction there. It would be the clearest though, as the fields of the signature/transaction are explained on Lockstacks, while doing it through `blockstack-cli` would imply either extra details in this document, or leaving readers without an understanding on what the fields are and why they need them. An extra point is the check on Lockstacks that doesn't allow you to stack unless you have the minimum amount in your wallet.

@kenrogers could you please give your opinion on this as well?

kenrogers: If the goal of the quickstart is getting up and running with a signer as quickly as possible, I think we can use the CLI here with a brief description of the commands they need to run and link people to the other methods of stacking if they want more options/details

wileyj: i think i will recant my suggestion here. only because using the cli would require a bit more documentation. it's not as simple apparently as just running the cli with some commands. basically, the pox address needs to be converted to a hash in order to use the cli and that would be out of scope for this document.

guides-and-tutorials/running-a-signer/signer-quickstart.md
aldur: Optimizing for making it as easy as possible for users, I think using a sub shell to get the value of `stacks_private_key` might be confusing, even though it's a great idea:

I'd do something in the middle:

```bash
# Set environment variables
AUTH_TOKEN="$(openssl rand -base64 12)"  # Used for signer-node authentication
PRIVATE_KEY="$(jq -r  '.privateKey' test.json)" # or privateKey from Step 1, this is the signer's private key

# Now create the signer's configuration file
cat <<EOF> ~/stacks-signer/signer-config.toml
node_host = "127.0.0.1:20443"
endpoint = "127.0.0.1:30000"
network = "testnet"
db_path = "$HOME/stacks-signer/data/signer.sqlite"
auth_password = "$AUTH_TOKEN"
stacks_private_key = "$PRIVATE_KEY"
metrics_endpoint = "127.0.0.1:9154"
block_proposal_timeout_ms = 180000
EOF
```

wileyj: let's leave as-is for now then

wileyj (approved):

PR #1665, opened 2024-10-24
fix: naka testnet 3.0 height

hstove:
There was an issue with setting it to 1000 so we changed it to 1900


wileyj (approved):

wileyj:
@hstove looks fine to me, but will need some conflicts resolved

PR #1672, opened 2024-11-06
fix: Suggest fall-back signers instead of active redundancy

aldur:
## Description

Reduce potential network impacts by recommending Signers to have a fallback instance ready, instead of active redundancy.

### DIFF
--- a/guides-and-tutorials/running-a-signer/best-practices-to-run-a-signer.md
+++ b/guides-and-tutorials/running-a-signer/best-practices-to-run-a-signer.md
@@ -50,6 +50,8 @@ tl;dr: avoid single point of failures, introduce redundancy, monitor things.
 
 - Upgrading one Signer instance at the time.
 - Test the update on one instance and, if successful, proceed to the others.
+- While your Signer is offline for upgrades, it won't sign any blocks. Ensure
+  that the downtime is as short as possible.
 
 ### Diversified hosting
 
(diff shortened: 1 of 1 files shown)


guides-and-tutorials/running-a-signer/best-practices-to-run-a-signer.md
```diff
@@ -10,15 +10,6 @@ The following best practices suggest how to create a resilient setup for running
 tl;dr: avoid single point of failures, introduce redundancy, monitor things.
 {% endhint %}
 
-### Run redundant Signer instances
```
friedger, line 13: Add a comment in Monitor your signer: "Ensure that only one instance is using the same private key"

guides-and-tutorials/running-a-signer/best-practices-to-run-a-signer.md
```diff
@@ -57,7 +48,7 @@ without interruption.
 
 ### Upgrade procedures
 
-- Upgrading one Signer instance (of those running redundantly) at the time.
+- Upgrading one Signer instance at the time.
```
friedger, line 51: What does that mean? Maybe 
"During upgrade, your signer is offline and won't sign any blocks. Ensure that the downtime is as short as possible."

guides-and-tutorials/running-a-signer/best-practices-to-run-a-signer.md
aldur: This doesn't fit in the "monitor" section, but I agree we might need to make it more explicit.

wileyj (approved):

PR #1676, opened 2024-11-27
fix: sBTC Signer: Point to testnet config and small fixes

aldur:
We now have testnet config / readme. This points interested readers theree.


kenrogers (approved):
LGTM

PR #1714, opened 2025-04-14
removing node_modules

wileyj:
`node_modules` is no longer needed after moving to a different platform. 
this PR removes these outdated files and *should* deploy successfully.


BowTiedRadone (approved):
This PR makes sense. Only `node_modules` folder is touched. LGTM 🚀

PR #1743, opened 2025-11-04
updated image src paths for moved files

eric-stacks:
## Description

in the previous PR merged, some files were moved to a new Space folder but the image src paths were not adjusted. this PR fixes that.


kenny-stacks (approved):
lgtm

PR #1744, opened 2025-11-06
feat: use @stacks/clarinet-sdk

hugo-stacks:
## Description

Use `@stacks/clarinet-sdk` instead of the hirosystems one


BowTiedRadone (dismissed):
LGTM!

repos-to-monitor.txt
```diff
@@ -7,6 +7,6 @@ stacks-network/clarity-wasm
 stacks-sbtc/sbtc
 
 # Hiro Systems Repositories
```
radubahmata: Now this heading is no longer accurate.

hugo-stacks: Right, good catch

eric-stacks (approved):
lgtm

PR #1745, opened 2025-11-07
fixed pdf src paths that didn't auto update in Gitbook

eric-stacks:
## Description

Moved the sBTC Audits page into a new grouping in the Gitbook UI, but the pdf src paths didn't auto update. This PR fixes those paths.


kenny-stacks (approved):
lgtm

PR #1747, opened 2025-11-11
moved clarinet/stacksjs guides from Reference to Build nav

eric-stacks:
As per the first milestone of the dev-related revamps, moving the Clarinet/Stacks.js guides from the Reference to Build nav will make the Build section more intuitive.

Moving these over first will make it much easier to also adjust the flow of these guides.

Clarinet/Stacks.js "actual reference" pages were left in the Reference nav.


kenny-stacks (approved):
lgtm

PR #1749, opened 2025-11-26
Rendezvous docs refinements and fixes

BowTiedRadone:
This PR refines the Rendezvous reference page and fixes two broken links.

### DIFF
--- a/docs/reference/rendezvous/reference.md
+++ b/docs/reference/rendezvous/reference.md
@@ -1,3 +1,9 @@
+---
+description: >-
+  Complete reference for Rendezvous CLI commands, configuration options, and
+  advanced usage.
+---
+
 # Rendezvous Reference
 
 This reference explains how to use Rendezvous in different situations. By the end, you'll know when and how to use its features effectively.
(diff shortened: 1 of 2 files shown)


docs/reference/rendezvous/reference.md
```diff
@@ -1,39 +1,13 @@
...
+description: >-
+  Complete reference for Rendezvous CLI commands, configuration options, and
+  advanced usage.
+---
+
 # Rendezvous Reference
 
 This reference explains how to use Rendezvous in different situations. By the end, you'll know when and how to use its features effectively.
 
-## What's Inside
```
wileyj, line 5: if we can leave this here, i think it would be helpful. @eric-stackslabs is it possible to link to a section like in generic markdown?

wileyj (commented):
lgtm pending the navigation section comment

docs/reference/rendezvous/reference.md
eric-stacks: yes, the way it is now works on the live docs

wileyj: @BowTiedRadone can you re-add this section? unless you feel strongly that it should be removed, i think it would be useful to keep.

radubahmata: There's already a ToC on the right of the page, felt that having an extra one is noisy. But now that I checked on a mobile device as well, I think you're right. Will re-add.

radubahmata: Done, dropped the last commit and amended the previous one 🙌

eric-stacks (approved):
lgtm

PR #1755, opened 2025-12-19
Fix the glob for allowed claude tools

kenny-stacks:
Small change to fix the glob that tells claude what tools it is allowed to use


ECBSJ (approved):
lgtm

eric-stacks (approved):
lgtm

PR #1836, opened 2026-02-25
feat: add a guide on running a node behind a proxy

### LINKED ISSUE #1762: Document how to run a node behind a proxy
Use case - running a public node unbounded is generally not the best idea. 
i propose adding a documentation on how to perform some basic rate-limits with off the shelf services like nginx or haproxy that a node operator may use if they want to run a node publicly.

simone-stacks:
## Summary                                                                                                                                                                                                                                                                         
  - Adds a new guide covering how to run a Stacks node behind a reverse proxy with rate limiting
  - Includes production-tested configurations for both HAProxy and Nginx
  - Updates the Docker guide with minor tips

## Additional info
Closes #1762 
Closes https://github.com/stx-labs/core-epics/issues/110

  ## Checklist
  - [ ] Review HAProxy and Nginx configuration examples for accuracy


docs/operate/SUMMARY.md
```diff
@@ -7,6 +7,7 @@
   * [Run a Node with Quicknode](readme/run-a-node-with-quicknode.md)
   * [Run a Bitcoin Node](readme/run-a-bitcoin-node.md)
   * [Run a Pruned Bitcoin Node](readme/run-a-pruned-bitcoin-node.md)
```
wileyj, line 9: this file seems to be missing/deleted. intentional? this is a guide that's probably worth keeping, even if moves folders

docs/operate/run-a-node/run-a-bitcoin-node.md
wileyj, line 1: Any specific reason this doc is deleted? it's definitely relevant info for stacks node operators.

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -0,0 +1,278 @@
+# Run a Node Behind a Proxy
+
+{% hint style="warning" %}
+Running a publicly accessible node exposes your infrastructure to the open internet. The configurations below are starting points, not complete security solutions. **Do this at your own risk.** You are responsible for securing and maintaining your own infrastructure.
```
wileyj:
```suggestion
Running a publicly accessible node exposes your infrastructure to the open internet. The proxy configurations below are starting points, not complete security solutions. **Do this at your own risk.** You are responsible for securing and maintaining your own infrastructure.
```

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -0,0 +1,278 @@
...
+{% endhint %}
+
+If you plan to run a Stacks node with publicly accessible RPC endpoints, it is strongly recommended to place the node behind a reverse proxy with rate limiting. Without rate limiting, a public node can be overwhelmed by excessive requests, leading to degraded performance or denial of service.
```
wileyj:
```suggestion
If you plan to run a Stacks node with publicly accessible RPC endpoints, it is strongly recommended at a minimum to place the node behind a reverse proxy with rate limiting. Without rate limiting, a public node can be overwhelmed by excessive requests, leading to degraded performance or denial of service.
```

`at a minimum` - i though to include something about a firewall, but that's out of scope here.

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -0,0 +1,278 @@
...
+
+This guide provides minimal, production-tested configurations for two popular reverse proxies. **Choose one — you do not need both:**
+
+- [**Nginx**](#nginx) — simpler configuration, widely known, good baseline rate limiting.
```
wileyj:
```suggestion
- [**Nginx**](#nginx) — simpler configuration, widely used, good baseline rate limiting.
```

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -0,0 +1,278 @@
...
+
+{% hint style="info" %}
+The **P2P ports** (20444, 8333) use custom binary protocols for peer-to-peer communication, not HTTP. You can leave them open directly to the network. The proxy configurations below focus on the **RPC/API ports** which serve HTTP traffic and are the primary target for abuse.
+
+**Optional:** P2P ports can also benefit from rate-limiting. While unlikely, a denial-of-service attack could flood the P2P port so the node only communicates with malicious peers. Adding connection-rate limits on P2P ports won't hurt and provides an extra layer of protection.
+{% endhint %}
+
+## Configure the Stacks node
+
+Before setting up the proxy, configure your Stacks node so its RPC endpoint is not directly reachable from the public internet. The proxy will be the only public-facing service.
```
wileyj: I think a short explanation here is good. something like `(ex: rpc_bind = "127.0.0.1:30443")`


```suggestion
Before setting up the proxy, configure your Stacks node so its RPC endpoint is not directly reachable from the public internet (i.e. for stacks-node configuration -`rpc_bind = "127.0.0.1:30443"`). The proxy will be the only public-facing service.
```
edit: updated port to match port used later

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -0,0 +1,278 @@
...
+
+The format is `host_ip:host_port:container_port`. The node inside the container keeps its default ports — only the **host** side changes. Offset host ports (`30443`, `33999`) are necessary because the proxy already occupies the standard ports (`20443`, `3999`) on the host. Binding to `127.0.0.1` ensures the container ports are only reachable from the host (where the proxy runs), not from the public internet. P2P is published directly on the standard port `20444`.
+
+{% hint style="info" %}
+Inter-container communication (e.g. the API receiving events from the blockchain node) uses Docker's internal network and service names, not published host ports. These port mapping changes do not affect container-to-container traffic.
+{% endhint %}
+
+## Nginx
+
+Nginx can serve as a reverse proxy with rate limiting using the `limit_req` module. The configuration below rate-limits the Stacks RPC and Stacks API endpoints.
```
wileyj, line 83: I think it's worth adding text here defining *what* the rate limit being set to, and what the burst option does in this context.

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -0,0 +1,278 @@
...
+sudo ufw allow 20444/tcp           # Stacks P2P (direct)
+sudo ufw allow 8333/tcp            # Bitcoin P2P (direct)
+sudo ufw enable
+```
+
+{% endcode %}
+
+{% hint style="warning" %}
+**Docker users:** Docker manipulates `iptables` directly and bypasses UFW rules. If your node runs in Docker, bind container ports to `127.0.0.1` explicitly (e.g. `-p 127.0.0.1:20443:20443`) or use the `DOCKER-USER` iptables chain to enforce restrictions. See the [Docker documentation](https://docs.docker.com/engine/network/packet-filtering-firewalls/) for details.
+{% endhint %}
```
wileyj, line 289: I feel like this doesn't add anything to this document and could be removed. 
however, i think there is a need for a firewall doc somewhere - but in terms of a proxy doc, it makes things a little more unclear - do users also need a firewall? and if so, there are many many ways to accomplish it (it really depends on how the hardware is running - a local host could use ufw, but iptables is also more configurable, amongst other options. then, consider that most operators will be using some cloud provider which all do firewalling differently). 

i think if you want to keep this section, i would keep it incredibly simple - mention that a firewall is beneficial and then link to a few common provider's docs on how to enable it.

docs/operate/readme/run-a-node-with-docker.md
```diff
@@ -55,6 +72,15 @@ Start the docker-compose services for the chosen network:
 ```
 {% endcode %}
 
+{% hint style="warning" %}
+**macOS users:** The `manage.sh` script uses GNU sed syntax that is incompatible with macOS BSD sed. If you encounter `sed: bad flag in substitute command` errors, install GNU sed and prepend it to your PATH:
+
+```bash
+brew install gnu-sed
+PATH="/opt/homebrew/opt/gnu-sed/libexec/gnubin:$PATH" ./manage.sh -n <network> -a start
```
wileyj: wouldn't it be better to persist this change?

i think this suggestion is better suited for the stacks-blockchain-docker repo rather than the docs repo

docs/operate/readme/run-a-node-with-docker.md
```diff
@@ -23,6 +23,23 @@ MacOS with an ARM (M-series chip) processor is NOT recommended
...
+# If persistent-data already exists, move its contents first
+mv persistent-data/mainnet /mnt/stacks-data/mainnet
+rmdir persistent-data
+
+# Create the symlink
+ln -s /mnt/stacks-data persistent-data
+```
+{% endcode %}
+
+The Docker volumes write to `persistent-data/mainnet/`, so the symlink redirects all chainstate, database, and event data to the external disk transparently.
```
wileyj:
```suggestion
The Docker volumes write to `persistent-data/<network>/`, so the symlink redirects all chainstate, database, and event data to the external disk transparently.
```

small nit

docs/operate/.gitbook.yaml
```diff
@@ -4,8 +4,8 @@ redirects:
 
   # nodes-and-miners to run-a-node redirects
   guides-and-tutorials/nodes-and-miners: README.md
-  guides-and-tutorials/nodes-and-miners/run-a-node-with-docker: run-a-node/run-a-node-with-docker.md
-  guides-and-tutorials/nodes-and-miners/run-a-node-with-digital-ocean: run-a-node/run-a-node-with-digital-ocean.md
+  guides-and-tutorials/nodes-and-miners/run-a-node-with-docker: readme/run-a-node-with-docker.md
+  guides-and-tutorials/nodes-and-miners/run-a-node-with-digital-ocean: readme/run-a-node-with-digital-ocean.md
```
wileyj, line 8: the linked file is being deleted in this PR. intentional?

wileyj: i think i see what's going on. hmmmm

the `run-a-node` folder is deprecated, and the content exists in new locations, correct?

i think you'll want to double-check all links in this PR to ensure they point to the correct files @simone-stacks - i've noticed several in this PR that appear to be incorrect.

docs/operate/SUMMARY.md
simone-stacks: I removed the whole run-a-node folder because it was a duplicate of the readme folder after discussing this with @eric-stacks, I don't know why it was duplicated
The preview in this PR seems fine tho: https://docs.stacks.co/operate/~/revisions/BbSyVUkoVyZFoU0aVJk9/readme/run-a-node-behind-a-proxy
But worth double checking

docs/operate/run-a-node/run-a-bitcoin-node.md
simone-stacks: Same as https://github.com/stacks-network/docs/pull/1836#discussion_r2888920775

docs/operate/readme/run-a-node-behind-a-proxy.md
simone-stacks: Absolutely! crc: [dac52a3](https://github.com/stacks-network/docs/pull/1836/commits/dac52a3543c9694a4c2ed645395d0a9e2675488f)

docs/operate/readme/run-a-node-behind-a-proxy.md
simone-stacks: Yes it makese sense, I kept the section but made it generic in this commit [dac52a3](https://github.com/stacks-network/docs/pull/1836/commits/dac52a3543c9694a4c2ed645395d0a9e2675488f)

docs/operate/readme/run-a-node-with-docker.md
simone-stacks: Removed it! crc: [dac52a3](https://github.com/stacks-network/docs/pull/1836/commits/dac52a3543c9694a4c2ed645395d0a9e2675488f)

docs/operate/.gitbook.yaml
simone-stacks: I double checked and all links seem to be pointing to the readme folder, also, in the preview https://docs.stacks.co/operate/~/revisions/BbSyVUkoVyZFoU0aVJk9/readme/run-a-node-behind-a-proxy everything seems to be correct, but it might be worth double checking @eric-stacks

simone-stacks:
In commit [651c964](https://github.com/stacks-network/docs/pull/1836/commits/651c9647c876f439f7312942ac8b63aa8cbdc443) I added a few improvements:
- In the ports table, Bitcoin RPC listed `8332` but HAProxy was binding to `18332` (testnet). Aligned to mainnet default `8332` and updated the table to show both variants.
- All three HAProxy frontends share one `Abuse` table, meaning a client that trips the rate limit on any service gets blocked from all of them. Added a comment explaining this behavior.
- Added `proxy_set_header` directives for Nginx so the Stacks node sees real client IPs in its logs instead of 127.0.0.1 (helps with debugging and log correlation)

Do you think those changes are correct @wileyj?

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -1,27 +1,27 @@
...
-| Stacks P2P  | 20444        | TCP      | No              |
-| Stacks API  | 3999         | HTTP     | Yes, if running |
-| Bitcoin RPC | 8332         | HTTP     | Yes, if exposed |
-| Bitcoin P2P | 8333         | TCP      | No              |
+| Service     | Default Port                     | Protocol | Proxy?          |
+| ----------- | -------------------------------- | -------- | --------------- |
+| Stacks RPC  | 20443                            | HTTP     | Yes             |
+| Stacks P2P  | 20444                            | TCP      | No              |
+| Stacks API  | 3999                             | HTTP     | Yes, if running |
+| Bitcoin RPC | 8332 (mainnet) / 18332 (testnet) | HTTP     | Yes, if exposed |
```
wileyj: I don't think this is necessary in the context of this doc - stacks testnet doesn't use bitcoin testnet anymore (it uses a regtest network because of testnet miner griefing a while back that caused numerous chain stalls). 
it's not common for a node operator on testnet to run a bitcoin regtest node (it does require some additional config vs a standard bitcoin setup). 

but, if you feel it's worth keeping - then i would also mention the p2p testnet port as well.

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -220,7 +238,7 @@ backend stacks_api_back
 # Bitcoin RPC (optional, if you expose it)
 # -------------------------------------------
 frontend btc_rpc
-    bind *:18332
+    bind *:8332
```
wileyj, line 241: good catch, i missed that!

docs/operate/readme/run-a-node-behind-a-proxy.md
```diff
@@ -0,0 +1,289 @@
...
+
+{% endcode %}
+
+{% hint style="info" %}
+**How the abuse table works:** HAProxy tracks each client IP's HTTP request rate. When a client exceeds the threshold (e.g. 25 HTTP requests in 10 seconds), its `gpc0` counter is incremented and all subsequent requests from that IP are denied with HTTP 429. The stick-table entry expires after 30 minutes, lifting the block automatically.
+{% endhint %}
+
+## Firewall considerations
+
+A host-level firewall adds defense in depth: only the proxy's listening ports and the P2P ports should be reachable from the public internet, while the node's RPC stays accessible only via the proxy (localhost). How you configure this depends on your environment — cloud providers, bare-metal hosts, and container setups all handle firewalling differently.
```
wileyj:
```suggestion
Additionally, a host-level firewall adds defense in depth: only the proxy's listening ports and the P2P ports should be reachable from the public internet, while the node's RPC stays accessible only via the proxy (localhost). How you configure this depends on your environment — cloud providers, bare-metal hosts, and container setups all handle firewalling differently.
```

Small qualifier suggestion that i think makes it more clear that a firewall is a good idea alongside a proxy. ack or nack, wording is also currently fine if you prefer it.

docs/operate/readme/run-a-node-behind-a-proxy.md
wileyj: to be clear, only the mentioned testnet port - the space formatting is fine

docs/operate/readme/run-a-node-behind-a-proxy.md
wileyj: looks good! might be worth @eric-stacks to create an issue to have a more in-depth firewall guide in the future?

docs/operate/readme/run-a-node-behind-a-proxy.md
simone-stacks: Oh I didn't know it, removed in https://github.com/stacks-network/docs/pull/1836/commits/8d689f3c87dc97a71aa9db006d538af7eb65f331

eric-stacks:
@simone-stacks just confirming if all the changes were resolved? looks like its good to go?

PR #1862, opened 2026-05-28
chore: remove .ds_store

hugo-stacks:
## Description
Remove root .DS_Store and add it to gitignore


cuevasm (approved):
Makes sense - this action can't be done from Gitbook either so we'll need another Approver 🙏

Update: Nevermind on the additional reviewer!

PR #1863, opened 2026-05-28
feat: add clarinet simnet-only docs

hugo-stacks:
## Description

Document the new Simnet-only Clarity code deployment in Clarinet


alexis-stacks (approved):
Changes look fine to me.

PR #1868, opened 2026-06-30
feat: auto-update from stacks-core

brice-stacks:
This adds a cronjob that will auto-generate the Clarity docs from the stacks-core repo, which is the source of truth. Requires PR stacks-network/stacks-core#7367 in that repo to add the notes to some elements.


brice-stacks:
@cuevasm this is my initial attempt to automate this. We'll have to check if it actually looks correct in the published site.

brice-stacks:
I think [the preview](https://docs.stacks.co/reference/~/revisions/swQTSLD6WulC5xzYi5wv/clarity/keywords) looks good!

cuevasm (approved):
Thanks Brice, preview looks solid so let's push this and see how things look live and especially at the next update.

PR #1869, opened 2026-06-30
feat: add latest testnet epoch heights

brice-stacks:


cuevasm (approved):
Confirming expected update of block heights in various places throughout

PR #1873, opened 2026-07-22
feat: update post-condition docs

brice-stacks:
Add updates from SIP-040 and SIP-045.


alexis-stacks (dismissed):
Approved. Caveat: I'm relying on the technical content being correct, but the docs changes look fine to me.

alexis-stacks (approved):
Rubber stamped again.

PR #1877, opened 2026-07-29
fix: miner wallet setup for mandatory `wallet_name` + Bitcoin Core >= 25 support

francesco-stacks:
## Description

Companion to stacks-network/stacks-core#7433: burnchain.wallet_name is now mandatory for miners, and the node loads (never creates) the wallet.

The miner guides now provision a watch-only descriptor wallet (`disable_private_keys=true`) and import the public `combo(...)` descriptor of the miner key: the stacks node signs its bitcoin transactions itself, so no private key is stored in bitcoind. This replaces the previous createwallet + importprivkey flow, which is broken on Bitcoin Core ≥ 29.

Every documented command was executed against the latest minor of every Core major from 25 through 31, and `stacks-core` now has an integration test https://github.com/stacks-network/stacks-core/pull/7433/changes/8692f3ecbaa97de18656da8419fd73563d7bf647  that pins the current guide commands for each new major version of bitcoin.


brice-stacks (approved):
lgtm. Thanks!

PR #1878, opened 2026-08-03
chore: update testnet configs for pox-5

rafa-stacks:
Updates the `Config.toml` Stacks node configs for testnet so users can follow the new pox-5 network.


brice-stacks (approved):
lgtm

PR #1879, opened 2026-08-06
fix: update seed nodes for testnet configurations

rafa-stacks:
Also add remaining funding accounts


brice-stacks (approved):
Looks good. We may need similar updates in the sample configs in stacks-core as well.

PR #1881, opened 2026-08-19
fix: use canonical sbtc testnet address

hugo-stacks:
On previous Testnet, we were using a dummy sip-010 token as sBTC (deployed by Hiro, linked to the old faucet)

On the new PoX-5 testnet, the canonical sBTC contract is preferred


mitchell-stacks (approved):
Makes sense, thanks for the update!

cuevasm (approved):
Second approval

PR #1882, opened 2026-08-20
fix: update testnet configs

brice-stacks:
Updated for the latest testnet.


rafa-stacks (approved):
Thanks @brice-stacks !

cuevasm (approved):
Awesome!

